PR #15, opened 2018-06-13
[issue-2] Handle exception in emitter and receiver

### LINKED ISSUE #2: Enforce error handling
_As a Shunting Yard user_
_I'd like to have errors handled and monitored properly_
_So I can make sure the service won't go down upon exceptions_

## Overview

Shunting Yard is meant to be a 24/7 service and as such it must do a correct handling of exceptional conditions:

Emitter side:
* What happens if marshaling fails?
* Post message in the topic?
* How do we notify about these errors if we care at all?

Receiver side:
* What happens if Circus Train replication fail?
* What happens if unmarshaling fails?
* How do we notify about these errors if we care at all?

## Acceptance Criteria

Emitter:
* Add metrics if possible: successes and failures. Log all errors using the keywords `Error`, `ShuntingYard` and `Emitter` so they can be analyzed easily.

Receiver:
*  Add metrics: ingestion successes and failures. Log all errors using the keywords `Error`, `ShuntingYard` and `Receiver` so they can be analyzed easily.

ddcprg:
Fixes #2


shunting-yard-common/src/main/java/com/hotels/shunting/yard/common/emitter/AbstractMetaStoreEventListener.java
```diff
@@ -45,9 +45,21 @@
...
 import com.hotels.shunting.yard.common.messaging.MessageTask;
 import com.hotels.shunting.yard.common.messaging.MessageTaskFactory;
+import com.hotels.shunting.yard.common.metrics.MetricsConstant;
+import com.hotels.shunting.yard.common.metrics.MetricsHelper;
 
 public abstract class AbstractMetaStoreEventListener extends MetaStoreEventListener {
-  private static final Logger LOG = LoggerFactory.getLogger(AbstractMetaStoreEventListener.class);
+  private static final Logger log = LoggerFactory.getLogger(AbstractMetaStoreEventListener.class);
+
+  static void success() {
```
massdosage: What's the reason these two methods are static?

shunting-yard-common/src/main/java/com/hotels/shunting/yard/common/emitter/WrappingMessageTask.java
```diff
@@ -0,0 +1,55 @@
...
+
+  public WrappingMessageTask(MessageTask task) {
+    this.task = task;
+  }
+
+  @Override
+  public void run() {
+    try {
+      task.run();
+      AbstractMetaStoreEventListener.success();
```
massdosage: Aah, that's why they're static. I guess an alternative to this would be for run to return a success/failure return value (or throw an exception) and the caller could call success/error based on that. Then this MessageTask wouldn't be coupled to the EventListener.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/app/ReplicationRunner.java
```diff
@@ -32,14 +32,26 @@
...
 import com.hotels.shunting.yard.common.receiver.ShuntingYardMetaStoreEventListener;
 
 @Component
 class ReplicationRunner implements ApplicationRunner, ExitCodeGenerator {
-  private static final Logger LOG = LoggerFactory.getLogger(ReplicationRunner.class);
+  private static final Logger log = LoggerFactory.getLogger(ReplicationRunner.class);
 
-  private ShuntingYardMetaStoreEventListener listener;
-  private MessageReader messageReader;
+  static void success() {
```
massdosage: Do these two need to be static?

shunting-yard-common/src/main/java/com/hotels/shunting/yard/common/emitter/AbstractMetaStoreEventListener.java
ddcprg: they don't depend in the class state

shunting-yard-common/src/main/java/com/hotels/shunting/yard/common/emitter/WrappingMessageTask.java
ddcprg: this runs on a separate thread

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/app/ReplicationRunner.java
ddcprg: they don't depend on the class itself

shunting-yard-common/src/main/java/com/hotels/shunting/yard/common/emitter/WrappingMessageTask.java
massdosage: OK, but it could return a Future to whatever is running it? I'm not saying we do this but the coupling and the call to a static method on an Abstract class looks weird.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/app/ReplicationRunner.java
massdosage: Then perhaps they shouldn't be here ;) Alternatively they could be private and non-static, what's the point of making something static if its never going to be called statically?

ddcprg: you need to look at how the bytecode is generated

massdosage: We should never need to look at the byte code unless there's some huge performance issue that requires it.

shunting-yard-common/src/main/java/com/hotels/shunting/yard/common/emitter/WrappingMessageTask.java
ddcprg: well, this class was supposed to be inside the abstract class but then I had to make things public to allow testing so I moved it to it's own file. I don't see what's weird about it, it's like when you call static methods of the logging API in you class, perhaps the name of the class?

ddcprg: I can move the static methods to an utility class but then utility classes are strange. I just want to use the same logger that the listener uses

massdosage: What did you need to make public? Wouldn't default level access with a "@visibleForTesting" annotation have worked if you wanted to keep things in one class?

massdosage: Make them protected then? Easier to test also if you don't have static stuff which is hard to mock etc. Might not be that relevant here. I guess another alternative would be to pass a non-static result listener object in that you could call methods on, that would be easy to mock if necessary. I can live with what's here too for now, I just think the risk of having static methods is that anybody can call them at any time even if they are completely unrelated to the class. This is OK for standard utility functionality like logging which you want to make "globabally" accessible. The problem here is that I could in some completely unrelated code call AbstractMetaStoreEventListener.success() and cause side effects. If you're OK with this then fine.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/app/ReplicationRunner.java
massdosage: Since these aren't being used in a static manner let's make them private and non-static for now.

shunting-yard-common/src/main/java/com/hotels/shunting/yard/common/emitter/WrappingMessageTask.java
ddcprg: no code is fool proof

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/app/ReplicationRunner.java
ddcprg: they are used in a static manner, I can move them to an utility class if this is an issue

massdosage: I only see them being used on lines 96 and 98 and these both have access to the methods non-statically don't they? Are the methods being called from elsewhere?

shunting-yard-common/src/main/java/com/hotels/shunting/yard/common/emitter/EmitterUtils.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+package com.hotels.shunting.yard.common.emitter;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.hotels.shunting.yard.common.metrics.MetricsConstant;
+import com.hotels.shunting.yard.common.metrics.MetricsHelper;
+
+public final class EmitterUtils {
```
massdosage: Since these methods are only being called in one place why not just move the MetricsHelper.incrementCounter(MetricsConstant.EMITTER_SUCCESSES); calls to where success()/erorr() is currently being called? Then we don't need this class or ReceiverUtils at all?

shunting-yard-common/src/main/java/com/hotels/shunting/yard/common/emitter/AbstractMetaStoreEventListener.java
```diff
@@ -81,44 +83,72 @@ private Message withPayload(SerializableListenerEvent event) throws MetaExceptio
 
   @Override
   public void onCreateTable(CreateTableEvent tableEvent) throws MetaException {
-    LOG.info("Create table event received");
-    executorService.submit(message(withPayload(serializableListenerEventFactory.create(tableEvent))));
+    log.info("Create table event received");
+    try {
```
massdosage: An Aspect could be a slightly nicer way to do this but perhaps not so relevant for only 7 methods, also given the issues we had with getting Aspects to work last time we tried using them.

shunting-yard-common/src/main/java/com/hotels/shunting/yard/common/emitter/WrappingMessageTask.java
```diff
@@ -0,0 +1,58 @@
...
+    try {
+      task.run();
+      success();
+    } catch (Exception e) {
+      error(e);
+    }
+  }
+
+  @Override
+  public boolean equals(Object obj) {
```
massdosage: Add a comment saying this is only here for testing?

shunting-yard-common/src/main/java/com/hotels/shunting/yard/common/emitter/AbstractMetaStoreEventListener.java
```diff
@@ -81,44 +83,72 @@ private Message withPayload(SerializableListenerEvent event) throws MetaExceptio
...
+      executorService.submit(message(withPayload(serializableListenerEventFactory.create(tableEvent))));
+    } catch (Exception e) {
+      error(e);
```
massdosage: Test that throwing an exception -> is caught and "errored"

shunting-yard-emitter/shunting-yard-emitter-kinesis/src/test/java/com/hotels/shunting/yard/emitter/kinesis/listener/KinesisMetaStoreEventListenerTest.java
```diff
@@ -97,7 +98,8 @@ public void onCreateTable() throws Exception {
     when(serializableEvent.getTableName()).thenReturn(TABLE);
     when(serializableListenerEventFactory.create(event)).thenReturn(serializableEvent);
     listener.onCreateTable(event);
-    verify(executorService).submit(messageTask);
+    verify(executorService).submit(captor.capture());
```
massdosage: Since this is covered by tests for Abstract class you can probably just verify that submit() was called with *something* and then wrapper class can be made package protected.

shunting-yard-emitter/shunting-yard-emitter-sqs/src/test/java/com/hotels/shunting/yard/emitter/sqs/listener/SqsMetaStoreEventListenerTest.java
```diff
@@ -97,7 +98,8 @@ public void onCreateTable() throws Exception {
     when(serializableEvent.getTableName()).thenReturn(TABLE);
     when(serializableListenerEventFactory.create(event)).thenReturn(serializableEvent);
     listener.onCreateTable(event);
-    verify(executorService).submit(messageTask);
+    verify(executorService).submit(captor.capture());
```
massdosage: Same here and others...

massdosage (changes requested):
Looks good, some small changes

shunting-yard-emitter/shunting-yard-emitter-sqs/src/test/java/com/hotels/shunting/yard/emitter/sqs/listener/SqsMetaStoreEventListenerTest.java
ddcprg: done

shunting-yard-emitter/shunting-yard-emitter-kinesis/src/test/java/com/hotels/shunting/yard/emitter/kinesis/listener/KinesisMetaStoreEventListenerTest.java
ddcprg: done

shunting-yard-common/src/main/java/com/hotels/shunting/yard/common/emitter/AbstractMetaStoreEventListener.java
ddcprg: done

shunting-yard-common/src/main/java/com/hotels/shunting/yard/common/emitter/WrappingMessageTask.java
ddcprg: done

ddcprg:
I may bump `platform-bom` in this PR

massdosage (approved):

shunting-yard-emitter/shunting-yard-emitter-kafka/src/test/java/com/hotels/shunting/yard/emitter/kafka/listener/KafkaMetaStoreEventListenerTest.java
```diff
@@ -97,7 +106,8 @@ public void onCreateTable() throws Exception {
     when(serializableEvent.getTableName()).thenReturn(TABLE);
     when(serializableListenerEventFactory.create(event)).thenReturn(serializableEvent);
     listener.onCreateTable(event);
-    verify(executorService).submit(messageTask);
+    verify(executorService).submit(captor.capture());
```
patduin: Why are you using a `captor` if you are not asserting anything on it?

patduin: ah got it reading the comments made by Adrian now

ddcprg: leftover from last change I made, I'll remove them

patduin (approved):

Issue #6, opened 2018-06-05
Allow users to select the tables to be replicated


ddcprg:
_As a Shunting Yard user_
_I'd like to be able to select the tables to be replicated from a source metastore_
_So I can only replicate those that are of interest for the analysts/applications_

## Acceptance criteria
User must be able to select tables from on of the pre-configured source metastores and select the target metastores where the table must be replicated.

ddcprg:
Note description mentions that there might be more than one source and more than one replica so this issue depends on another which I'll be creating shortly

massdosage:
We should try allow Regular Expressions for choosing the tables (https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html) in a similar fashion to what we do in the Waggle Dance whitelisting.

PR #19, opened 2018-12-07
Added Table Filtering

abhimanyugupta07:
refer #6

### DIFF
--- a/shunting-yard-binary/src/main/scripts/replicator.sh
+++ b/shunting-yard-binary/src/main/scripts/replicator.sh
@@ -33,8 +33,8 @@ fi
 LIBFB303_JAR=`ls $HIVE_LIB/libfb303-*.jar | tr '\n' ':'`
 
 SHUNTING_YARD_LIBS=$SHUNTING_YARD_HOME/lib/*\
-:$HIVE_LIB/hive-exec.jar\
-:$HIVE_LIB/hive-metastore.jar\
+:$HIVE_LIB/hive-exec*.jar\
+:$HIVE_LIB/hive-metastore*.jar\
 :$LIBFB303_JAR\
 :$HIVE_CONF_PATH/
 
--- a/shunting-yard-binary/src/main/scripts/replicator.sh
+++ b/shunting-yard-binary/src/main/scripts/replicator.sh
@@ -33,8 +33,8 @@ fi
 LIBFB303_JAR=`ls $HIVE_LIB/libfb303-*.jar | tr '\n' ':'`
 
 SHUNTING_YARD_LIBS=$SHUNTING_YARD_HOME/lib/*\
-:$HIVE_LIB/hive-exec*.jar\
-:$HIVE_LIB/hive-metastore*.jar\
+:$HIVE_LIB/hive-exec.jar\
+:$HIVE_LIB/hive-metastore.jar\
 :$LIBFB303_JAR\
 :$HIVE_CONF_PATH/
 
--- a/shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/TargetReplication.java
+++ b/shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/TargetReplication.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2016-2018 Expedia Inc.
+ * Copyright (C) 2016-2019 Expedia Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,6 +15,7 @@
  */
 package com.hotels.shunting.yard.replicator.exec.conf;
 
+import java.util.ArrayList;
 import java.util.List;
 
 import org.springframework.boot.context.properties.ConfigurationProperties;
@@ -23,7 +24,7 @@ import org.springframework.context.annotation.Configuration;
 @Configuration
 @ConfigurationProperties(prefix = "target-replication")
 public class TargetReplication {
-  private List<String> tableNames;
+  private List<String> tableNames = new ArrayList<>();
 
   public List<String> getTableNames() {
     return tableNames;
--- a/shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/TargetReplication.java
+++ b/shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/TargetReplication.java
@@ -22,7 +22,7 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.context.annotation.Configuration;
 
 @Configuration
-@ConfigurationProperties(prefix = "target-replication")
+@ConfigurationProperties(prefix = "source-table-filter")
 public class TargetReplication {
   private List<String> tableNames = new ArrayList<>();
 
--- a/shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/TargetReplication.java
+++ b/shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/TargetReplication.java
@@ -1,0 +1,36 @@
+/**
+ * Copyright (C) 2016-2018 Expedia Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hotels.shunting.yard.replicator.exec.conf;
+
+import java.util.List;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ConfigurationProperties(prefix = "target-replication")
+public class TargetReplication {
+  private List<String> tableNames;
+
+  public List<String> getTableNames() {
+    return tableNames;
+  }
+
+  public void setTableNames(List<String> tableNames) {
+    this.tableNames = tableNames;
+  }
+
+}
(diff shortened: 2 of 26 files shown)


shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/MessageReaderAdapter.java
```diff
@@ -65,6 +71,10 @@ private MetaStoreEvent map(ListenerEvent listenerEvent) {
 
     EventType eventType = listenerEvent.getEventType();
 
+    if (!tableSelector.canProcess(listenerEvent)) {
+      return builder.build();
```
patduin: This still returns an event for the table we don't want to process. How is this going to prevent the table replication from happening?

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/context/CommonBeans.java
```diff
@@ -149,16 +151,22 @@ MetaStoreEventAggregator eventAggregator() {
...
       ApiarySqsMessageDeserializer sqsMessageSerDe,
       EventReceiverConfiguration messageReaderConfig,
-      SourceCatalog sourceCatalog) {
+      SourceCatalog sourceCatalog,
+      TableSelector tableSelector) {
     MessageReaderFactory messaReaderFactory = MessageReaderFactory
         .newInstance(messageReaderConfig.getMessageReaderFactoryClass());
     MessageReader messageReader = messaReaderFactory.newInstance(replicaHiveConf, sqsMessageSerDe);
-    return new MessageReaderAdapter(messageReader, sourceCatalog.getHiveMetastoreUris());
+    return new MessageReaderAdapter(messageReader, sourceCatalog.getHiveMetastoreUris(), tableSelector);
```
patduin: I think if possible the Adapter should not know about the filtering of tables it should get ListenerEvents that are valid for processing. Perhaps  we need to look at a `FilteringMessageReader` that filters out the stream of event. Can be a decarator on whaterever MessageReader implementation we use currently?

shunting-yard-binary/src/main/scripts/replicator.sh
```diff
@@ -33,8 +33,8 @@ fi
 LIBFB303_JAR=`ls $HIVE_LIB/libfb303-*.jar | tr '\n' ':'`
 
 SHUNTING_YARD_LIBS=$SHUNTING_YARD_HOME/lib/*\
-:$HIVE_LIB/hive-exec.jar\
-:$HIVE_LIB/hive-metastore.jar\
+:$HIVE_LIB/hive-exec*.jar\
```
massdosage: Why do we need this? I'm guessing we have version numbers in some situations? (just curious what those situations are).

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/TargetReplication.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package com.hotels.shunting.yard.replicator.exec.conf;
+
+import java.util.List;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ConfigurationProperties(prefix = "target-replication")
```
massdosage: This matches the layout in the YAML right? Where it makes sense I think we should re-use the same names as we use in Circus Train, so I was thinking more like something like this:

```
    table-replications:
      - source-table: 
          database-name: foo
          table-name: bar
```
i.e. in the YAML file you can provide a list of tables to replicate. We could also allow regex to be used for both the db and table name and then compile this into a list of potential matches.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/context/CommonBeans.java
massdosage: I like the idea of using a decorator to filter out the tables we're not interested in.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/receiver/TableSelector.java
```diff
@@ -0,0 +1,37 @@
...
+public class TableSelector {
+
+  private final TargetReplication targetReplication;
+
+  public TableSelector(TargetReplication targetReplication) {
+    this.targetReplication = targetReplication;
+  }
+
+  public boolean canProcess(ListenerEvent listenerEvent) {
+    List<String> tableNames = targetReplication.getTableNames();
```
massdosage: You can do this in the constructor and keep a reference to the tables to match rather than the `targetReplication` itself (or perhaps even better, the constructor should just take a list and not be coupled to the `targetReplication` at all).

shunting-yard-replicator/src/test/java/com/hotels/shunting/yard/replicator/exec/messaging/FilteringMessageReaderTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+  private static final String DB_NAME = "test_db";
+  private static final String TABLE_NAME1 = "test_table1";
+  private static final String TABLE_NAME2 = "test_table2";
+  private static final String TABLE_NAME3 = "test_table3";
+
+  private @Mock ListenerEvent listenerEvent1;
+  private @Mock ListenerEvent listenerEvent2;
+  private @Mock ListenerEvent listenerEvent3;
+  private @Mock SqsMessageReader delegate;
```
patduin: `SqsMessageReader` -> `MessageReader` No need to mock the concrete class

shunting-yard-replicator/src/test/java/com/hotels/shunting/yard/replicator/exec/messaging/FilteringMessageReaderTest.java
```diff
@@ -0,0 +1,93 @@
...
+  @Before
+  public void init() {
+    when(listenerEvent1.getDbName()).thenReturn(DB_NAME);
+    when(listenerEvent1.getTableName()).thenReturn(TABLE_NAME1);
+    when(listenerEvent2.getDbName()).thenReturn(DB_NAME);
+    when(listenerEvent2.getTableName()).thenReturn(TABLE_NAME2);
+    when(listenerEvent3.getDbName()).thenReturn(DB_NAME);
+    when(listenerEvent3.getTableName()).thenReturn(TABLE_NAME3);
+
+    when(delegate.hasNext()).thenReturn(true);
```
patduin: should probably mock this in a way that it respects the MessageReader interface. So hasNext should return true if there is a next otherwise false

shunting-yard-replicator/src/test/java/com/hotels/shunting/yard/replicator/exec/messaging/FilteringMessageReaderTest.java
```diff
@@ -0,0 +1,93 @@
...
+  public void selectFirstAndThirdEventButSkipSecond() {
+    when(tableSelector.canProcess(listenerEvent1)).thenReturn(true);
+    when(tableSelector.canProcess(listenerEvent2)).thenReturn(false);
+    when(tableSelector.canProcess(listenerEvent3)).thenReturn(true);
+
+    List<ListenerEvent> expectedEventList = new ArrayList<>();
+    filteringMessageReader = new FilteringMessageReader(delegate, tableSelector);
+    ListenerEvent event = null;
+
+    for (int i = 0; i < 2; i++) {
```
patduin: this is a bit weird perhaps
 ```
assertThat(filteringMessageReader.hasNext(), is(true));
assertThat(expectedEventList.get(0).getDbName()).isEqualTo(DB_NAME);
assertThat(expectedEventList.get(0).getTableName()).isEqualTo(TABLE_NAME1);
assertThat(filteringMessageReader.hasNext(), is(true));
assertThat(expectedEventList.get(1).getDbName()).isEqualTo(DB_NAME);
assertThat(expectedEventList.get(1).getTableName()).isEqualTo(TABLE_NAME2);
assertThat((filteringMessageReader.hasNext(), is(false));
```

File: shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/FilteringMessageReader.java [hunk 9]
```diff
@@ -0,0 +1,56 @@
...
+
+  @Override
+  public boolean hasNext() {
+    while (delegate.hasNext()) {
+      current = delegate.next();
+      if (tableSelector.canProcess(current)) {
+        return true;
+      }
+    }
+    return false;
```
patduin: Probably best to set current to null before returning

shunting-yard-replicator/src/test/java/com/hotels/shunting/yard/replicator/exec/messaging/FilteringMessageReaderTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+    assertThat(filteringMessageReader.hasNext(), is(false));
+  }
+
+  @Test
+  public void emptyDelegateReader() {
+    when(delegate.hasNext()).thenReturn(false);
+    filteringMessageReader = new FilteringMessageReader(delegate, tableSelector);
+
+    assertThat(filteringMessageReader.hasNext(), is(false));
```
patduin: assert that the next() == null?

patduin:
Add a line to the changelog please and once we settle on how the config should look we maybe need to add a short section in the README on how to use it the feature.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/FilteringMessageReader.java
```diff
@@ -32,6 +33,12 @@ public FilteringMessageReader(MessageReader delegate, TableSelector tableSelecto
     this.tableSelector = tableSelector;
   }
 
+  /**
```
patduin: remove this, this is not true. This class has nothing to do with the SQSMessageReader. This filtering will work for any implementation.
The `MessageReader` extends `Iterator` so hasNext will not be removed in the future. We just follow the iterator contract. See https://docs.oracle.com/javase/7/docs/api/java/util/Iterator.html.

Concerning the SQSMessageReader I understand return true is confusing but it does not break the contract. It doesn't say you cannot wait indefinetly for next to return a new value it just says: "return true if the iterator has more elements"

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/FilteringMessageReader.java
```diff
@@ -0,0 +1,64 @@
...
+        return true;
+      }
+    }
+    current = null;
+    return false;
+  }
+
+  @Override
+  public ListenerEvent next() {
+    return current;
```
patduin: Looking at the Iterator contract in more depth we should do:
```
if (current == null) {
  throw new NoSuchElementException();
}
return current;
```

PR #20, opened 2019-01-06
Table filter with better behaving "hasNext()"

massdosage:
This turned into a much bigger change than I had anticipated but the basic idea is that since the `FilteringMessageReader` can filter messages out we'd rather return these as empty Optionals than block and wait for something that passes the filter to arrive. While making these changes I also removed the two implementations of `Iterator` since these weren't true iterators any more and were abusing the contract (e.g. always returning true for hasNext(), throwing Exceptions for unimplemented remove() methods etc.). I then noticed that the main loop controlling the `ReplicationRunner` could be improved by firing it off in a separate thread and running while true

### DIFF
--- a/shunting-yard-common/src/main/java/com/hotels/shunting/yard/common/metrics/MetricsConstant.java
+++ b/shunting-yard-common/src/main/java/com/hotels/shunting/yard/common/metrics/MetricsConstant.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2016-2018 Expedia Inc.
+ * Copyright (C) 2016-2019 Expedia Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,10 +17,8 @@ package com.hotels.shunting.yard.common.metrics;
 
 public final class MetricsConstant {
 
-  public static final String EMITTER_FAILURES = "shunting_yard_emitter_failures";
-  public static final String EMITTER_SUCCESSES = "shunting_yard_emitter_successes";
-
   public static final String RECEIVER_FAILURES = "shunting_yard_receiver_failures";
+  public static final String RECEIVER_EMPTY = "shunting_yard_receiver_empty";
   public static final String RECEIVER_SUCCESSES = "shunting_yard_receiver_successes";
 
   private MetricsConstant() {}
--- a/shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/MetaStoreEventReplication.java
+++ b/shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/MetaStoreEventReplication.java
@@ -98,11 +98,6 @@ public class MetaStoreEventReplication {
   }
 
   @Bean
-  public TaskExecutor taskExecutor() {
-    return new SimpleAsyncTaskExecutor();
-  }
-
-  @Bean
   MetricRegistry runningMetricRegistry() {
     return new MetricRegistry();
   }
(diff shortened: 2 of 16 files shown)


shunting-yard-common/src/main/java/com/hotels/shunting/yard/common/metrics/MetricsConstant.java
```diff
@@ -17,10 +17,8 @@
 
 public final class MetricsConstant {
 
-  public static final String EMITTER_FAILURES = "shunting_yard_emitter_failures";
```
massdosage: We no longer have emitters so I cleaned these up.

shunting-yard-common/src/main/java/com/hotels/shunting/yard/common/metrics/MetricsConstant.java
```diff
@@ -17,10 +17,8 @@
...
-  public static final String EMITTER_SUCCESSES = "shunting_yard_emitter_successes";
-
   public static final String RECEIVER_FAILURES = "shunting_yard_receiver_failures";
+  public static final String RECEIVER_EMPTY = "shunting_yard_receiver_empty";
```
massdosage: I figured we may as well count these so we can plot how many times they happen, should probably "never" happen since the `SqsMessageReader` sits in a loop waiting for messages to arrive before returning the call to `next()`.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/MetaStoreEventReplication.java
```diff
@@ -48,15 +50,16 @@ public static void main(String[] args) throws Exception {
 
     int exitCode = -1;
     try {
-      exitCode = SpringApplication.exit(new SpringApplicationBuilder(MetaStoreEventReplication.class)
```
massdosage: This is just a formatting change, not sure why.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/MetaStoreEventReplication.java
```diff
@@ -94,6 +97,11 @@ private static void logVersionInfo() {
     LOG.info("MetaStoreEventReplication");
   }
 
+  @Bean
+  public TaskExecutor taskExecutor() {
```
massdosage: Added this so we have a simple way to fire off a thread to run the main app.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/app/ReplicationRunner.java
```diff
@@ -32,41 +35,63 @@
...
+      TaskExecutor executor) {
     this.listener = listener;
     this.eventReader = eventReader;
+    this.executor = executor;
   }
 
   @Override
   public void run(ApplicationArguments args) {
-    while (eventReader.hasNext()) {
+    executor.execute(this);
```
massdosage: This effectively runs this class in a separate thread.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/app/ReplicationRunner.java
```diff
@@ -32,41 +35,63 @@
...
+  }
+
+  @Override
+  public int getExitCode() {
+    return 0;
+  }
+
+  public void stop() {
```
massdosage: Will probably only ever be used by the unit test as Spring will handle stopping the thread if the application context is closed some other way. We will want to revisit this stop/start logic and should use a fancier executor that can stop the receiver from receiving new messages but wait until any messages in flight are completed before shutting down. For now this does the job and shouldn't be any worse than what we had before.

massdosage: Yeah, I agree it's not ideal but I'd rather not spend the time on it now. I think we need to sort out overall how this app runs and particularly how it shuts down so should sort all of that out at the same time and can improve this then. I'll make a note to create a ticket for this.

PR #28, opened 2019-02-01
Graphite config for Shunting Yard

abhimanyugupta07:
Adding a new config file to be passed directly to internal CT

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,22 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## TBD
+## [0.0.8] TBD
+### Added
+* Support for passing certain configurations like Graphite directly to internal CT instance using `--ct-config` argument.
+
+## [0.0.7] 2019-01-28
+### Added
+* log4j.xml to be packaged with Shunting-Yard Binary.
+
+## [0.0.6] 2019-01-22
 ### Changed
 * Upgraded `hive` to 2.3.4 (was 2.3.0).
+* ContainerCredentialsProvider to a more robust EC2ContainerCredentialsProvider.
+
+## [0.0.5] 2019-01-14
+### Added
+* Support to read AWS Credentials from within the Elastic Container Service Task using ContainerCredentialsProvider.
 
 ## [0.0.4] 2019-01-08
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,27 +7,27 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) a
 ### Added
 * Support for passing certain configurations like Graphite directly to internal CT instance using `--ct-config` argument.
 
-## [0.0.7] 2019-01-28
+## [0.0.7] - 2019-01-28
 ### Added
 * log4j.xml to be packaged with Shunting-Yard Binary.
 
-## [0.0.6] 2019-01-22
+## [0.0.6] - 2019-01-22
 ### Changed
 * Upgraded `hive` to 2.3.4 (was 2.3.0).
 * ContainerCredentialsProvider to a more robust EC2ContainerCredentialsProvider.
 
-## [0.0.5] 2019-01-14
+## [0.0.5] - 2019-01-14
 ### Added
 * Support to read AWS Credentials from within the Elastic Container Service Task using ContainerCredentialsProvider.
 
-## [0.0.4] 2019-01-08
+## [0.0.4] - 2019-01-08
 ### Changed
 * Refactored project to remove checkstyle and findbugs warnings, which does not impact functionality.
 * Upgraded `hotels-oss-parent` to 2.3.5 (was 2.1.0).
 ### Added
 * Support for selecting the tables to be replicated [#6](https://github.com/HotelsDotCom/shunting-yard/issues/6).
 
-## [0.0.3] 2018-11-01
+## [0.0.3] - 2018-11-01
 ### Changed
 * Enforce exception handling [#2](https://github.com/HotelsDotCom/shunting-yard/issues/2).
 
@@ -38,10 +38,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) a
 ### Removed
 * Receivers & emitters for Kinesis and Kafka.
 
-## [0.0.2] 2018-06-06 - not ready for production
+## [0.0.2] - 2018-06-06 - not ready for production
 ### Changed
 * Generate a fat emitter JAR per emitter implementation [#13](https://github.com/HotelsDotCom/shunting-yard/issues/13).
 
-## [0.0.1] 2018-06-05 - not ready for production
+## [0.0.1] - 2018-06-05 - not ready for production
 ### Added
 * First release.
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Create a `.yml` file under `$SHUNTING_YARD_HOME/conf/` with the correct settings
     
 ### Graphite Metrics
 
-Graphite configurations can be passed to Shunting Yard using an optional `--ct-config` argument which takes a YAML file and passes it directly to internal Circus Train instance. Refer to the Circus-Train README [here](https://github.com/HotelsDotCom/circus-train#graphite) for more details.
+Graphite configurations can be passed to Shunting Yard using an optional `--ct-config` argument which takes a YAML file and passes it directly to internal Circus Train instance. Refer to the [Circus Train README](https://github.com/HotelsDotCom/circus-train#graphite) for more details.
 
 #### Sample ct-config.yml:
 
--- a/README.md
+++ b/README.md
@@ -39,6 +39,18 @@ Copy the file _shunting-yard-binary/target/shunting-yard-binary-\<version\>-bin.
 Create a `.yml` file under `$SHUNTING_YARD_HOME/conf/` with the correct settings to talk to the messaging infrastructure. You can find guidelines in the file _shunting-yard-minimal.yml.template_. Once this is done you are all set up to run the service:
 
     $SHUNTING_YARD_HOME/bin/replicator.sh --config=$SHUNTING_YARD_HOME/conf/<my-config>.yml
+    
+### Graphite Metrics
+
+Graphite configurations can be passed to Shunting Yard using an optional `--ct-config` argument which takes a YAML file and passes it directly to internal Circus Train instance. Refer to the Circus-Train README [here](https://github.com/HotelsDotCom/circus-train#graphite) for more details.
+
+Sample ct-config.yml:
+
+    graphite:
+      enabled: true
+      host: localhost:2003
+      namespace: com.company.shuntingyard
+      prefix: dev
 
 # Legal
 This project is available under the [Apache 2.0 License](http://www.apache.org/licenses/LICENSE-2.0.html).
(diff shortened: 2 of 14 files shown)


shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/context/CommonBeans.java
```diff
@@ -72,10 +73,14 @@
   private static final Logger LOG = LoggerFactory.getLogger(CommonBeans.class);
 
   @Bean
-  Configuration baseConfiguration(@Value("${instance.workspace}") String workspace) {
+  Configuration baseConfiguration(
+      @Value("${instance.workspace}") String workspace,
+      @Value("${ct-config}") String circusTrainConfigLocation) {
     checkNotNull(workspace, "instance.workspace is required");
+    checkNotNull(circusTrainConfigLocation, "ct-config is required");
```
massdosage: Being pedantic here since we're the only users of Shunting Yard up to now, but this would be a backwards incompatible change for end users right? We should always think carefully about these things, if we really want to do them that's fine they just require a major version bump and a clear line in the CHANGELOG describing the change and what users need to do. 

Could there be a situation where users don't want to provide a ct config so we could keep this optional?

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/receiver/ContextFactory.java
```diff
@@ -93,6 +94,7 @@ private String workOutReplicaLocation(String databaseName, String tableName) {
   }
 
   public Context createContext(MetaStoreEvent event) {
+    String circusTrainConfigLocation = PropertyUtils.stringProperty(conf, CT_CONFIG);
```
massdosage: Should we maybe turn this into a File and check that it exists? (uncertain where we should do this but what happens if the file doesn't exist? what will break and when?)

abhimanyugupta07:
This is still a WIP. I am not done with these changes. Please hold off on reviewing these for now. I just created the PR to get some initial feedback on my approach.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/context/CommonBeans.java
patduin: I think this doesn't have to be required. The basic config is made on the fly, this "common" one contains extras (grafana, maybe sns). So can be optional.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/context/CommonBeans.java
```diff
@@ -72,10 +73,16 @@
   private static final Logger LOG = LoggerFactory.getLogger(CommonBeans.class);
 
   @Bean
-  Configuration baseConfiguration(@Value("${instance.workspace}") String workspace) {
+  Configuration baseConfiguration(
+      @Value("${instance.workspace}") String workspace,
+      @Value("${ct-config:#{null}}") String circusTrainConfigLocation) {
```
abhimanyugupta07: `@Value("${ct-config:#{null}}")` is the syntax for getting null value if `ct-config` is not provided.

Spring throws:

`java.lang.IllegalArgumentException: Could not resolve placeholder 'ct-config' in value "${ct-config}"`

if we don't set the value as null explicitly.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/receiver/ContextFactory.java
abhimanyugupta07: The argument is optional now.

shunting-yard-replicator/src/test/java/com/hotels/shunting/yard/replicator/exec/receiver/ContextFactoryTest.java
```diff
@@ -164,6 +167,32 @@ public void createContextForNotExistingReplicaTable() throws Exception {
     assertThat(replication.getReplicaTable().getDatabaseName()).isEqualTo(DATABASE);
     assertThat(replication.getReplicaTable().getTableName()).isEqualTo(TABLE);
     assertThat(replication.getReplicaTable().getTableLocation()).isEqualTo(REPLICA_DATABASE_LOCATION + "/" + TABLE);
+    assertThat(context.getCircusTrainConfigLocation()).isNull();
+  }
+
+  @Test
+  public void ctConfigLocationGetsAddedToTheContextWhenProvided() {
+    conf.set(CT_CONFIG.key(), "ct-config.yml");
```
abhimanyugupta07: Adding CT_CONFIG to the Configuration object and testing that it gets set in the Context.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/context/CommonBeans.java
abhimanyugupta07: The argument is optional now.

abhimanyugupta07:
> Probably good to get into the habit of a CHANGELOG for this project and doing proper version numbers etc.
> 
> Where are we going to put the new CT config file that we want to use? In the docker or TF side of things?

These changes were made by me as and when I found a better way of doing things to test in Docker / ECS. So, I have updated the CHANGELOG for all of them together.

CT config file will be handled in the Docker side for Automatic deployment. In other cases, user can provide it optionally. README is updated to reflect that.

README.md
```diff
@@ -39,6 +39,18 @@ Copy the file _shunting-yard-binary/target/shunting-yard-binary-\<version\>-bin.
...
     $SHUNTING_YARD_HOME/bin/replicator.sh --config=$SHUNTING_YARD_HOME/conf/<my-config>.yml
+    
+### Graphite Metrics
+
+Graphite configurations can be passed to Shunting Yard using an optional `--ct-config` argument which takes a YAML file and passes it directly to internal Circus Train instance. Refer to the Circus-Train README [here](https://github.com/HotelsDotCom/circus-train#graphite) for more details.
+
+#### Sample ct-config.yml:
+
+    graphite:
+      enabled: true
```
patduin: we don't have enabled as a property right in CT?
we derive it based on having a host/namespace/port.
I think there is a getter for enabled but no setter.

abhimanyugupta07: Yep. Removed.

CHANGELOG.md
```diff
@@ -3,9 +3,22 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## TBD
+## [0.0.8] TBD
+### Added
+* Support for passing certain configurations like Graphite directly to internal CT instance using `--ct-config` argument.
+
+## [0.0.7] 2019-01-28
```
massdosage: For future reference the keepachangelog.com format is `## [1.0.0] - 2017-06-20` (i.e. with a dash separator)

README.md
```diff
@@ -39,6 +39,17 @@ Copy the file _shunting-yard-binary/target/shunting-yard-binary-\<version\>-bin.
 Create a `.yml` file under `$SHUNTING_YARD_HOME/conf/` with the correct settings to talk to the messaging infrastructure. You can find guidelines in the file _shunting-yard-minimal.yml.template_. Once this is done you are all set up to run the service:
 
     $SHUNTING_YARD_HOME/bin/replicator.sh --config=$SHUNTING_YARD_HOME/conf/<my-config>.yml
+    
+### Graphite Metrics
+
+Graphite configurations can be passed to Shunting Yard using an optional `--ct-config` argument which takes a YAML file and passes it directly to internal Circus Train instance. Refer to the Circus-Train README [here](https://github.com/HotelsDotCom/circus-train#graphite) for more details.
```
massdosage: I think it will look better if the link is over the text "Circus Train README" and then remove the word "here" (also remove the dash between Circus and Train.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/receiver/ContextFactory.java
massdosage: OK, what happens if you set this but to a path that doesn't exist? Does something throw an error and if so what does it say?

abhimanyugupta07: I would want SY to fail to start. It will need to happen in commonBeans though as context gets created only on receipt of event. 

how does this look:

```
@Bean
  Configuration baseConfiguration(
      @Value("${instance.workspace}") String workspace,
      @Value("${ct-config:#{null}}") String circusTrainConfigLocation) {
    checkNotNull(workspace, "instance.workspace is required");
    Configuration baseConf = new Configuration();
    baseConf.set(WORKSPACE.key(), workspace);

    if (circusTrainConfigLocation != null) {
      validateCtConfigPath(circusTrainConfigLocation);
      baseConf.set(CT_CONFIG.key(), circusTrainConfigLocation);
    }
    return baseConf;
  }

  private void validateCtConfigPath(String circusTrainConfigLocation) {
    List<String> errors = new ArrayList<>();
    if (StringUtils.isBlank(circusTrainConfigLocation)) {
      errors.add("No config file was specified.");
    } else {
      for (String configFileString : Splitter.on(',').split(circusTrainConfigLocation)) {
        File configFile = new File(configFileString);
        if (!configFile.exists()) {
          errors.add("Config file " + configFileString + " does not exist.");
        } else if (!configFile.isFile()) {
          errors.add("Config file " + configFileString + " is a directory.");
        } else if (!configFile.canRead()) {
          errors.add("Config file " + configFileString + " cannot be read.");
        }
      }
    }
    if (!errors.isEmpty()) {
      throw new ConfigFileValidationException(errors);
    }
  }
```

massdosage: Looks good, you should be able to unit test most of those paths too.

abhimanyugupta07: Done. I have added a ConfigFileValidator to handle it in couple of places.

CHANGELOG.md
abhimanyugupta07: Fixed in whole file.

README.md
abhimanyugupta07: Done.

shunting-yard-replicator/src/test/java/com/hotels/shunting/yard/replicator/exec/receiver/ContextFactoryTest.java
```diff
@@ -164,6 +167,32 @@ public void createContextForNotExistingReplicaTable() throws Exception {
...
+    assertThat(context.getConfigLocation()).matches(actualWorkspacePattern + "/replication.yml");
+    verify(marshaller).marshall(eq(context.getConfigLocation()), circusTrainConfigCaptor.capture());
+    CircusTrainConfig circusTrainConfig = circusTrainConfigCaptor.getValue();
+    assertThat(circusTrainConfig.getSourceCatalog().getHiveMetastoreUris()).isEqualTo(SOURCE_METASTORE_URIS);
+    assertThat(circusTrainConfig.getReplicaCatalog().getHiveMetastoreUris()).isEqualTo(REPLICA_METASTORE_URIS);
+    assertThat(circusTrainConfig.getTableReplications()).hasSize(1);
+    TableReplication replication = circusTrainConfig.getTableReplications().get(0);
+    assertThat(replication.getReplicationMode()).isSameAs(ReplicationMode.FULL);
+    assertThat(replication.getSourceTable().getTableName()).isEqualTo(TABLE);
+    assertThat(replication.getSourceTable().getTableName()).isEqualTo(TABLE);
```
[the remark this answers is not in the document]
abhimanyugupta07: Good catch ! fixed along with this suggested refactoring.

PR #29, opened 2019-03-16
metadata-only-sync

abhimanyugupta07:
Changes for performing metadata only sync using shunting yard.


shunting-yard-common/src/test/java/com/hotels/shunting/yard/common/io/jackson/DeserializerWithJsonInputTest.java
```diff
@@ -61,71 +68,108 @@
...
   private final static String TEST_DB = "some_db";
   private final static String TEST_TABLE = "some_table";
+  private final static String TEST_TABLE_LOCATION = "s3://table_location";
+  private final static String OLD_TEST_TABLE_LOCATION = "s3://old_table_location";
+  private final static String PARTITION_LOCATION = "s3://table_location/partition_location";
+  private final static String OLD_PARTITION_LOCATION = "s3://table_location/old_partition_location";
+
+  @Before
+  public void init() throws IOException {
+    addPartitionEvent = IOUtils
```
patduin: We usually try to use: 
```
    <dependency>
      <groupId>fm.last.commons</groupId>
      <artifactId>lastcommons-test</artifactId>
      <version>5.2.1</version>
      <scope>test</scope>
    </dependency>
```
```
  @Rule
  public DataFolder dataFolder = new ClassDataFolder();
```
As that is a bit more flexible then using `getResource`. Arguably these files are not really resources but more like data for the test. Resources would be files that configure stuff, like `hibernate properties` or `log4j.xml`.

massdosage: :+1:

shunting-yard-common/src/test/resources/com/hotels/shunting/yard/common/io/jackson/drop_partition.json
```diff
@@ -0,0 +1,14 @@
...
+    \"tableName\":\"some_table\",
+    \"tableLocation\":\"s3://table_location\",
+    \"partitionKeys\":{\"col_1\": \"string\", \"col_2\": \"integer\",\"col_3\":\"string\"},
+    \"partitionValues\":[\"val_1\", \"val_2\", \"val_3\"],
+    \"partitionLocation\":\"s3://table_location/partition_location\",
+    \"tableParameters\":{
+        \"param_1\": \"val_1\",
+        \"param_2\": \"val_2\"
+    }
+}"
```
massdosage: Missing closing line break

shunting-yard-common/src/test/resources/com/hotels/shunting/yard/common/io/jackson/drop_table.json
```diff
@@ -0,0 +1,7 @@
+"Message" : "{
+    \"protocolVersion\":\"1.0\",
+    \"eventType\":\"DROP_TABLE\",
+    \"dbName\":\"some_db\",
+    \"tableName\":\"some_table\",
+    \"tableLocation\":\"s3://table_location\"
+    }"
```
massdosage: Ditto

shunting-yard-common/src/test/resources/com/hotels/shunting/yard/common/io/jackson/insert_table.json
```diff
@@ -0,0 +1,9 @@
+"Message" : "{
+    \"protocolVersion\":\"1.0\",
+    \"eventType\": \"INSERT\",
+    \"dbName\": \"some_db\",
+    \"tableName\":\"some_table\",
+    \"files\": [\"file:/a/b.txt\",\"file:/a/c.txt\"],
+    \"fileChecksums\":[\"123\",\"456\"],
+    \"partitionKeyValues\": {\"col_1\": \"val_1\",\"col_2\": \"val_2\",\"col_3\": \"val_3\"}
+    }"
```
massdosage: Ditto

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/MessageReaderAdapter.java
```diff
@@ -74,6 +76,10 @@ private MetaStoreEvent map(ListenerEvent listenerEvent) {
       break;
     case ALTER_PARTITION:
       AlterPartitionEvent alterPartition = (AlterPartitionEvent) listenerEvent;
+
+      if (alterPartition.getPartitionLocation() == alterPartition.getOldPartitionLocation()) {
```
massdosage: Any particular reason for using `==` over `.equals()` - I usually prefer the latter as it's more type safe (see https://stackoverflow.com/questions/513832/how-do-i-compare-strings-in-java). This probably works most of the time but there's no guarantee that the strings in question will interned and thus have equal object references.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/MessageReaderAdapter.java
```diff
@@ -88,6 +94,12 @@ private MetaStoreEvent map(ListenerEvent listenerEvent) {
       builder.partitionColumns(new ArrayList<>(insertTable.getPartitionKeyValues().keySet()));
       builder.partitionValues(new ArrayList<>(insertTable.getPartitionKeyValues().values()));
       break;
+    case ALTER_TABLE:
+      AlterTableEvent alterTable = (AlterTableEvent) listenerEvent;
+      if (alterTable.getTableLocation() == alterTable.getOldTableLocation()) {
```
massdosage: Same

shunting-yard-replicator/src/test/java/com/hotels/shunting/yard/replicator/exec/messaging/MessageReaderAdapterTest.java
```diff
@@ -57,19 +59,25 @@
...
 
   private MessageReaderAdapter messageReaderAdapter;
 
-  private @Mock AddPartitionEvent addApiaryPartitionEvent;
-  private @Mock AlterPartitionEvent alterApiaryPartitionEvent;
-  private @Mock DropPartitionEvent dropApiaryPartitionEvent;
-  private @Mock CreateTableEvent createApiaryTableEvent;
-  private @Mock InsertTableEvent insertApiaryTableEvent;
-  private @Mock DropTableEvent dropApiaryTableEvent;
+  private @Mock AddPartitionEvent apiaryAddPartitionEvent;
```
massdosage: The new names are better, nice.

shunting-yard-common/src/test/java/com/hotels/shunting/yard/common/io/jackson/DeserializerWithJsonInputTest.java
abhimanyugupta07: done.

shunting-yard-common/src/test/resources/com/hotels/shunting/yard/common/io/jackson/drop_partition.json
abhimanyugupta07: done.

shunting-yard-common/src/test/resources/com/hotels/shunting/yard/common/io/jackson/drop_table.json
abhimanyugupta07: done.

shunting-yard-common/src/test/resources/com/hotels/shunting/yard/common/io/jackson/insert_table.json
abhimanyugupta07: done.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/MessageReaderAdapter.java
abhimanyugupta07: Totally missed this. Fixed now.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/MessageReaderAdapter.java
abhimanyugupta07: fixed.

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [1.0.0] - TBD
+### Added
+* Support for handling new elements (location and table parameters) coming in Apiary events.
```
AnanaMJ: Should it be "**from** Apiary events"?

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@ All notable changes to this project will be documented in this file.
...
+* Support for performing metadata only sync in alter table and alter partitions event.
```
AnanaMJ: "alter table and alter partitions event**s**"

shunting-yard-common/src/test/java/com/hotels/shunting/yard/common/io/jackson/DeserializerWithJsonInputTest.java
```diff
@@ -61,71 +73,104 @@
...
+  public void init() throws IOException {
+    addPartitionEvent = new String(Files.readAllBytes(dataFolder.getFile("add_partition.json").toPath()))
+        .replace("\n", "");
+    alterPartitionEvent = new String(Files.readAllBytes(dataFolder.getFile("alter_partition.json").toPath()))
+        .replace("\n", "");
+    dropPartitionEvent = new String(Files.readAllBytes(dataFolder.getFile("drop_partition.json").toPath()))
+        .replace("\n", "");
+    createTableEvent = new String(Files.readAllBytes(dataFolder.getFile("create_table.json").toPath()))
+        .replace("\n", "");
+    insertEvent = new String(Files.readAllBytes(dataFolder.getFile("insert_table.json").toPath())).replace("\n", "");
```
AnanaMJ: All other variables are named <action>table/partitionEvent. Can this be named `insertTableEvent` please? :)

CHANGELOG.md
abhimanyugupta07: it should be. Fixed.

shunting-yard-common/src/test/java/com/hotels/shunting/yard/common/io/jackson/DeserializerWithJsonInputTest.java
abhimanyugupta07: Yeah. InsertTableEvent doesn't sound right to me :P

shunting-yard-common/src/test/java/com/hotels/shunting/yard/common/io/jackson/DeserializerWithJsonInputTest.java
```diff
@@ -61,71 +73,104 @@
...
   private final static String TEST_TABLE = "some_table";
+  private final static String TEST_TABLE_LOCATION = "s3://table_location";
+  private final static String OLD_TEST_TABLE_LOCATION = "s3://old_table_location";
+  private final static String PARTITION_LOCATION = "s3://table_location/partition_location";
+  private final static String OLD_PARTITION_LOCATION = "s3://table_location/old_partition_location";
+
+  @Before
+  public void init() throws IOException {
+    addPartitionEvent = new String(Files.readAllBytes(dataFolder.getFile("add_partition.json").toPath()))
+        .replace("\n", "");
```
patduin: why do you need to replace '\n'?

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/MessageReaderAdapter.java
```diff
@@ -74,6 +76,10 @@ private MetaStoreEvent map(ListenerEvent listenerEvent) {
       break;
     case ALTER_PARTITION:
       AlterPartitionEvent alterPartition = (AlterPartitionEvent) listenerEvent;
+
+      if (alterPartition.getPartitionLocation().equals(alterPartition.getOldPartitionLocation())) {
```
patduin: this `getPartitionLocation` can be null right?

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/MessageReaderAdapter.java
```diff
@@ -88,6 +94,12 @@ private MetaStoreEvent map(ListenerEvent listenerEvent) {
       builder.partitionColumns(new ArrayList<>(insertTable.getPartitionKeyValues().keySet()));
       builder.partitionValues(new ArrayList<>(insertTable.getPartitionKeyValues().values()));
       break;
+    case ALTER_TABLE:
+      AlterTableEvent alterTable = (AlterTableEvent) listenerEvent;
+      if (alterTable.getTableLocation().equals(alterTable.getOldTableLocation())) {
```
patduin: same here

shunting-yard-common/src/test/java/com/hotels/shunting/yard/common/io/jackson/DeserializerWithJsonInputTest.java
abhimanyugupta07: coz its not a valid json string with '\n' characters in it.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/MessageReaderAdapter.java
abhimanyugupta07: Theoretically its not possible because we have a null check on the partition here: https://github.com/ExpediaInc/apiary-extensions/blob/master/apiary-metastore-listener/src/main/java/com/expedia/apiary/extensions/metastore/listener/ApiarySnsListener.java#L169

I can't think of a situation where a partition is not on a location. Have you seen such a situation ?

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/MessageReaderAdapter.java
abhimanyugupta07: I am not sure. Can we have a table on a NULL location ?

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/MessageReaderAdapter.java
patduin: current situation, messages are sent without a location so they will be null

shunting-yard-common/src/test/java/com/hotels/shunting/yard/common/io/jackson/DeserializerWithJsonInputTest.java
massdosage: What error do you get (and from where) if you don't remove the '\n'? Just curious, feels like we should be able to handle human readable JSON, maybe there's just a setting we could use when we configure the mapper or something. If not then we'll have to live with this.

shunting-yard-common/src/test/java/com/hotels/shunting/yard/common/io/jackson/DeserializerWithJsonInputTest.java
```diff
@@ -61,71 +73,104 @@
...
   private final static String TEST_DB = "some_db";
   private final static String TEST_TABLE = "some_table";
+  private final static String TEST_TABLE_LOCATION = "s3://table_location";
+  private final static String OLD_TEST_TABLE_LOCATION = "s3://old_table_location";
+  private final static String PARTITION_LOCATION = "s3://table_location/partition_location";
+  private final static String OLD_PARTITION_LOCATION = "s3://table_location/old_partition_location";
+
+  @Before
+  public void init() throws IOException {
+    addPartitionEvent = new String(Files.readAllBytes(dataFolder.getFile("add_partition.json").toPath()))
```
massdosage: When you create a string from a byte[] as you're doing here, you should use the String constructor which allows you to specify the Charset of the file's contents so you don't get different behaviour if the platform default changes on different machines where this is run.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/MessageReaderAdapter.java
massdosage: If that's the case then we should use `Objects.equals()` here so we don't have to faff around with null checks.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/MessageReaderAdapter.java
massdosage: Also not sure but I guess if we use `Objects.equals` above then we might as well do the same here and then we're covered.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/MessageReaderAdapter.java
patduin: I just don't want to tie us to a listener version also deployment will be hard because current messages in the queue don't have a location. Will be easier if we gracefully handle it.

shunting-yard-common/src/test/java/com/hotels/shunting/yard/common/io/jackson/DeserializerWithJsonInputTest.java
abhimanyugupta07: Found this to fix this issue:

`mapper.configure(JsonParser.Feature.ALLOW_UNQUOTED_CONTROL_CHARS, true)`

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/MessageReaderAdapter.java
abhimanyugupta07: `Objects.equals` won't be enough for our case because we don't want to do a `METADATA_UPDATE` when `alterPartition.getPartitionLocation() == null`.

So, added an explicit null check before the `.equals()` check.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/MessageReaderAdapter.java
abhimanyugupta07: Same response as above.

shunting-yard-common/src/test/java/com/hotels/shunting/yard/common/io/jackson/DeserializerWithJsonInputTest.java
abhimanyugupta07: Good point. Changed it to the constructor with Charset set as UTF_8.

PR #30, opened 2019-03-19
Added tests for com.hotels.shunting.yard.common.messaging.Message

paulkennethkent:
I've analysed your code base and noticed that `shunting.yard.common.messaging.Message` is not fully tested.

I've written some tests for the functions in this class with the help of [Diffblue Cover](https://www.diffblue.com/). Hopefully, these tests should help you detect any regressions caused by future code changes. If you would find it useful to have additional tests written for this repository, I would be more than happy to look at other particular classes that you consider important.

### DIFF
--- a/shunting-yard-common/src/test/java/com/hotels/shunting/yard/common/messaging/MessageTest.java
+++ b/shunting-yard-common/src/test/java/com/hotels/shunting/yard/common/messaging/MessageTest.java
@@ -40,7 +40,7 @@ public class MessageTest {
   }
 
   @Test
-  public void nullTable() {
+  public void emptyTable() {
     Message.Builder objectUnderTest = createMessageBuilder("test_db_1", "",
         "foo", 1553013512L);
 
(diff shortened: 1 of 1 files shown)


shunting-yard-common/src/test/java/com/hotels/shunting/yard/common/messaging/MessageTest.java
```diff
@@ -0,0 +1,65 @@
...
+    objectUnderTest.payload("foo");
+    objectUnderTest.timestamp(1515585600000L);
+
+    assertThatExceptionOfType(IllegalArgumentException.class).isThrownBy(() -> objectUnderTest.build())
+        .withMessage("Parameter 'table' is required");
+  }
+  
+  @Test
+  public void testBuild(){
+    final Message.Builder objectUnderTest = Message.builder();
```
AnanaMJ: Since each test sets these every time:
```java
final Message.Builder objectUnderTest = Message.builder();
objectUnderTest.database("test_db");
objectUnderTest.timestamp(1515585600000L);
```
there could be a `@Setup` method with either the common code or a typical setup, and then each test could have fewer setup lines for `objectUnderTest`, which would emphasise more what parameter is causing an exception.

paulkennethkent:
Thanks for the feedback @AnanaMJ and @massdosage, I've made the changes you requested.

If you have any other projects or classes you would like me to have a look at then please let me know.

PR #27, opened 2019-01-30
Adding option to pass graphite config to Shunting Yard

abhimanyugupta07:
Adding option to pass graphite config to Shunting Yard which will be used to create CT config internally.

PR #31, opened 2019-03-25
Bdp 1477 sqs receiver

max-jacobs:
This PR extracts the `SqsMessageReader` and related classes as these will now be in Apiary.

[Here](https://github.com/ExpediaInc/apiary-extensions/pull/27) is the complementary PR for Apiary.


shunting-yard-binary/src/main/resources/shunting-yard-minimal.yml.template
```diff
@@ -9,6 +9,8 @@ event-receiver:
   configuration-properties:
     com.hotels.shunting.yard.event.receiver.sqs.queue: https://sqs.us-west-2.amazonaws.com/1234567/sqs-queue
     com.hotels.shunting.yard.event.receiver.sqs.wait.time.seconds: 15
+    # max.messages acceptable values: 1 to 10, default is 10. For more details, see https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ReceiveMessage.html
```
massdosage: Is this something you noticed was missing while going through this?

shunting-yard-receiver/shunting-yard-receiver-sqs/src/main/java/com/hotels/shunting/yard/receiver/sqs/messaging/SqsMessageReaderFactory.java
```diff
@@ -15,17 +15,36 @@
...
-import com.hotels.shunting.yard.common.messaging.MessageReader;
+import com.amazonaws.services.sqs.AmazonSQS;
+import com.amazonaws.services.sqs.AmazonSQSClientBuilder;
+
+import com.expedia.apiary.extensions.receiver.common.messaging.MessageReader;
+import com.expedia.apiary.extensions.receiver.sqs.messaging.SqsMessageReader;
+
 import com.hotels.shunting.yard.common.messaging.MessageReaderFactory;
 
 public class SqsMessageReaderFactory implements MessageReaderFactory {
```
massdosage: This looks "common" - isn't it? Just wondering why this wasn't pulled out?

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/AggregatingMetaStoreEventReader.java
```diff
@@ -48,11 +48,9 @@
       long maxExecTime = windowUnits.toMillis(window);
       long startTime = System.currentTimeMillis();
       while (startTime + maxExecTime > System.currentTimeMillis()) {
-        // TODO: since delegate.next() effectively blocks until some messages arrive, the below
```
massdosage: I think this TODO is still true as an issue and so we should probably keep the comment, even if not as a TODO but more as a warning.

massdosage (changes requested):
Looks mostly good, a few small comments, also needs a major version bump and a CHANGELOG.

shunting-yard-binary/src/main/resources/shunting-yard-minimal.yml.template
max-jacobs: This is something that I've added. `SqsMessageReader.Builder.withMaxMessages()` lets you configure the maximum number of messages returned when you poll a queue. So I've added the config property here, along with a short description of how to use it.

Do you think this is sufficient or should I document this elsewhere?

shunting-yard-receiver/shunting-yard-receiver-sqs/src/main/java/com/hotels/shunting/yard/receiver/sqs/messaging/SqsMessageReaderFactory.java
max-jacobs: Shunting Yard here uses the hadoop `Configuration` to supply the required queue, max messages, wait time config etc. I didn't think this would be common as Hkaas would load these from a spring config file, for example. Does that make sense?

shunting-yard-binary/src/main/resources/shunting-yard-minimal.yml.template
massdosage: This is fine, it's just that we're now mixing two very different things in one PR - a new feature and a big removal of code. I'd generally prefer these as two separate PRs, which is a bit more work but I think makes things much easier to review.

shunting-yard-receiver/shunting-yard-receiver-sqs/src/main/java/com/hotels/shunting/yard/receiver/sqs/messaging/SqsMessageReaderFactory.java
massdosage: OK, sure, that's fine. Let's leave it here then, if later on we see more commonality across different modules we can extract things further.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/AggregatingMetaStoreEventReader.java
max-jacobs: Yes this is still _kind of_ true. Previously, `SqsMessageReader.next()` would block until messages are returned:

`while (records == null || !records.hasNext()) { records = consumer.receiveMessage(...)... }`

now, however, it will only block for `waitTimeSeconds`, the amount of time to wait for messages to arrive, as I've removed the loop to allow whoever uses this library to implement this as they like. Happy to add this back in as a warning if you think it still warrants it?

shunting-yard-binary/src/main/resources/shunting-yard-minimal.yml.template
massdosage: OK, if we stick with this PR for now then we need a line in the changelog describing this new feature.

max-jacobs: I am happy to remove this instead to keep this PR clean, and we can add this in later on? I don't mind, up to you :)

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/AggregatingMetaStoreEventReader.java
massdosage: Ah, OK, so that's very hard to see since the change is in what is showing up as a new file in the other PR. Anyway, from what I can see that's done by setting the `withWaitTimeSeconds` on the `ReceiveMessageRequest` in `SqsMessageReader` which is in the other PR. I know we had some long discussions about how this all works a few months ago, when my brain has some capacity I'd like to think if there are any edge cases this change will have an impact on. Does the message reader return an empty list if the timeout occurs? So it will keep looping since an empty message list is returned but it won't block? 

As for the TODO comment, I guess the change means there could still be this problem if you provide a weird combination of values - like a `waitTimeSeconds` bigger than the `maxExecTime` but that's not our fault so we don't need to document it...

shunting-yard-binary/src/main/resources/shunting-yard-minimal.yml.template
massdosage: If it's not too much work and is possible to split them into two prs that would be nice - one does the move "as is" with hopefully no changes other than re-packaging and a few other small necessary changes and the other introduces this new feature?

max-jacobs: Done, removed from this PR

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/AggregatingMetaStoreEventReader.java
max-jacobs: Yes, from the aws docs [here]:(https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/sqs/model/ReceiveMessageRequest.html#getWaitTimeSeconds--) 

> If no messages are available and the wait time expires, the call returns successfully with an empty list of messages.

So if the `ReceiveMessageRequest` times out it will return an empty list, which will be returned as `Optional.empty()` by `SqsMessageReader`. 

`AggregatingMetaStoreEventReader` will then continue the loop to read (until `startTime + maxExecTime > System.currentTimeMillis()). So the block will only happen for the duration of `waitTimeSeconds` while waiting to receive the message.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/context/CommonBeans.java
```diff
@@ -149,13 +149,6 @@ MetaStoreEventDeserializer metaStoreEventSerDe(ObjectMapper objectMapper) {
     return new JsonMetaStoreEventDeserializer(objectMapper);
   }
 
-  @Bean
-  ApiarySqsMessageDeserializer sqsMessageSerDe(
```
abhimanyugupta07: There is also a bean for ObjectMapper in this class, can that be removed as well ?

max-jacobs: Good catch, `ObjectMapper` can be removed, as can the `MetaStoreEventDeserializer` bean

max-jacobs: Done

shunting-yard-receiver/shunting-yard-receiver-sqs/src/main/java/com/hotels/shunting/yard/receiver/sqs/messaging/SqsMessageReaderFactory.java
abhimanyugupta07: Actually this feature to read credentials and region from configuration is not used. I am thinking should we remove this and then we don't need the shunting-yard-receiver and receiver-sqs anymore.

shunting-yard-common/pom.xml
```diff
@@ -20,6 +20,13 @@
       <artifactId>commons-lang3</artifactId>
     </dependency>
 
+    <!-- Apiary -->
+    <dependency>
```
massdosage: Follow on work - let's see if we can remove this dependency from common to receiver-sqs

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) a
 * Support for handling new elements (location and table parameters) coming in from Apiary events.
 * Support for performing metadata only sync in alter table and alter partition event.
 
+### Changed
+* Extracted Apiary sqs receiver logic to library in [`apiary-extensions`](https://github.com/ExpediaInc/apiary-extensions).
```
massdosage: sqs -> SQS

massdosage: You should link to the full `apiary-receivers` location, you can put the full path in here just make sure we merge the other PR first.

massdosage (changes requested):
2 small things on the changelog, and major version bump in the poms please!

shunting-yard-common/pom.xml
```diff
@@ -20,6 +20,13 @@
...
+      <groupId>com.expedia.apiary</groupId>
```
abhimanyugupta07: My last suggestion would be, if its possible to create a variable for this version number in shunting-yard parent `pom.xml` so that it can be consistent across the project.

max-jacobs: Done!

max-jacobs:
@abhimanyugupta07 @patduin @massdosage I've addressed all comments and this is now ready for a final review before merging, thanks all

abhimanyugupta07 (approved):

massdosage (approved):
Approved, major version bump please (preferably now in the PR or later when releasing)

PR #32, opened 2019-03-25
Option to select target DB and table

abhimanyugupta07:
An option in the config file to specify target db and table for selected tables in shunting yard.

The configuration file will have a section like this:

The config will only apply if the table gets filtered through the `source-table-filter`:

```
table-replications:
  - source-table:
      database-name: bdp
      table-name: etl_daily_unique_visitors
    replica-table:
      database-name: test
      table-name: etl_daily_unique_visitors_1

### CODE REVIEW

File: shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/context/CommonBeans.java [hunk 1]
```diff
@@ -180,6 +184,18 @@ MessageReaderAdapter messageReaderAdapter(
     return new MessageReaderAdapter(filteringMessageReader, sourceCatalog.getHiveMetastoreUris());
   }
 
+  private Map<String, TableReplication> transformTableReplicationsToMap(TableReplications tableReplications) {
```
patduin: Maybe make an object to encapsulate this. `ShuntingYardTableReplications`? It can have some public look up methods that the ContextFactory can use.

abhimanyugupta07: done.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/receiver/ContextFactory.java
```diff
@@ -107,13 +115,26 @@ public Context createContext(MetaStoreEvent event) {
...
+
+    String replicaDatabaseName = event.getDatabaseName();
+    String replicaTableName = event.getTableName();
+
+    TableReplication tableReplication = shuntingYardTableReplications
+        .getTableReplicationsMap()
+        .get(String.join(".", event.getDatabaseName(), event.getTableName()));
+
+    if (tableReplication != null) {
+      replicaDatabaseName = tableReplication.getReplicaDatabaseName();
```
max-jacobs: Is the null check needed here? It's already checked in `ShuntingYardTableReplications` when the `tableReplications` are added to the map

max-jacobs: Ah I see, if its null we want to use the event to work out the location instead. Never mind! :)

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/receiver/ContextFactory.java
```diff
@@ -107,13 +115,26 @@ public Context createContext(MetaStoreEvent event) {
 
   private CircusTrainConfig generateConfiguration(MetaStoreEvent event) {
     String sourceMetaStoreUri = getSourceMetaStoreUri(event);
-    String replicaTableLocation = workOutReplicaLocation(event.getDatabaseName(), event.getTableName());
+
+    String replicaDatabaseName = event.getDatabaseName();
+    String replicaTableName = event.getTableName();
+
+    TableReplication tableReplication = shuntingYardTableReplications
+        .getTableReplicationsMap()
```
patduin: make the `getTableReplicationsMap` private and add a method `shuntingYardTableReplications.getTableReplication(String databaseName, String tableName)`

encapsulate all the things! + the joining the db and table name will be done in one class

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/receiver/ContextFactory.java
```diff
@@ -107,13 +115,26 @@ public Context createContext(MetaStoreEvent event) {
...
+        .get(String.join(".", event.getDatabaseName(), event.getTableName()));
+
+    if (tableReplication != null) {
+      replicaDatabaseName = tableReplication.getReplicaDatabaseName();
+      replicaTableName = tableReplication.getReplicaTableName();
```
patduin: you need to check if either db name or table name is actually set. Use could have set only one wanting to override for instance only the db name. You need to add tests for this as well

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/CircusTrainTableReplications.java
```diff
@@ -0,0 +1,25 @@
...
+package com.hotels.shunting.yard.replicator.exec.conf;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+import com.hotels.bdp.circustrain.api.conf.TableReplications;
+
+@Configuration
+@ConfigurationProperties(prefix = "table-replications")
+public class CircusTrainTableReplications extends TableReplications {}
```
patduin: If you run this do we not get errors because of the constraints e.g. `@NonNull`, `@TableReplicationFullReplicationModeConstraint` or do we not validate the config in SY?

patduin (changes requested):
added some comments

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/receiver/ContextFactory.java
abhimanyugupta07: That can't happen because the TableReplication class has a @NotNull constraint on SourceTable and ReplicaTable which in turn have @NotBlank constraints on DbName and TableName.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/receiver/ContextFactory.java
abhimanyugupta07: done

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/receiver/ContextFactory.java
abhimanyugupta07: Hold on for this. Investigating more.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ct/CircusTrainTableReplications.java
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+package com.hotels.shunting.yard.replicator.exec.conf.ct;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration("tableReplications")
+@ConfigurationProperties(prefix = "")
+public class CircusTrainTableReplications extends SyTableReplications {}
```
patduin: This class has no use anymore

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ct/SyReplicaTable.java
```diff
@@ -0,0 +1,40 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hotels.shunting.yard.replicator.exec.conf.ct;
+
+import org.hibernate.validator.constraints.NotBlank;
+
+public class SyReplicaTable {
```
patduin: I prefer `ReplicaTable`

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ct/SyReplicaTable.java
```diff
@@ -0,0 +1,40 @@
...
+  private @NotBlank String databaseName;
```
patduin: I think this can be blank or null or whatever, same for tableName

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ct/SyTableReplication.java
```diff
@@ -0,0 +1,59 @@
...
+ * limitations under the License.
+ */
+package com.hotels.shunting.yard.replicator.exec.conf.ct;
+
+import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
+
+import com.hotels.bdp.circustrain.api.conf.SourceTable;
+
+public class SyTableReplication {
```
patduin: `TableReplication`

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ct/SyTableReplication.java
```diff
@@ -0,0 +1,59 @@
...
+  private @Valid @NotNull SourceTable sourceTable;
+  private @Valid @NotNull SyReplicaTable replicaTable;
```
patduin: can be null

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ct/SyTableReplications.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package com.hotels.shunting.yard.replicator.exec.conf.ct;
+
+import java.util.List;
+
+import javax.validation.Valid;
+
+import org.hibernate.validator.constraints.NotEmpty;
+
+public class SyTableReplications {
```
patduin: `TableReplications`

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ct/SyTableReplications.java
```diff
@@ -0,0 +1,36 @@
...
+
+  private @Valid @NotEmpty List<SyTableReplication> tableReplications;
```
patduin: Can be empty that's ok, that means don't override anything right?

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ct/SyTableReplication.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+package com.hotels.shunting.yard.replicator.exec.conf.ct;
+
+import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
+
+import com.hotels.bdp.circustrain.api.conf.SourceTable;
+
+public class SyTableReplication {
+  private @Valid @NotNull SourceTable sourceTable;
```
patduin: I think Source table should also be copied. less confusing

shunting-yard-replicator/pom.xml
```diff
@@ -96,6 +97,16 @@
       <groupId>org.hibernate</groupId>
       <artifactId>hibernate-validator</artifactId>
     </dependency>
+    
+   <dependency>
```
patduin: why are these need?

shunting-yard-binary/src/main/resources/shunting-yard-minimal.yml.template
```diff
@@ -13,3 +13,10 @@ source-table-filter:
   table-names:
     - db1.table_1
     - db2.table_2
+table-replications:
```
patduin: Not sure if this is used in a test, but you probably still want the default where there are no table-replications in the template. I still think no override will be the most common use case.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/receiver/ContextFactory.java
abhimanyugupta07: OK. I see what you were saying. But I thought we decided that for now we won't support this usecase because we need to see how this feature will be used. It is possible, that a person would want to set a DB for all selected tables etc. So, for now, its going to be one db / table combination mapped to one db / table combination and the user needs to specify both.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/CircusTrainTableReplications.java
abhimanyugupta07: We do get errors and that is why CT conf classes need to be copied over to SY. It all looks ugly now :(

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ct/CircusTrainTableReplications.java
abhimanyugupta07: True. Removed.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ct/SyReplicaTable.java
abhimanyugupta07: This is SyReplicaTable because the CT's ReplicaTable is being used in the `CircusTrainConfig` class in SY. So, if I name it same as CT, it creates really ugly scenes in that class.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ct/SyReplicaTable.java
abhimanyugupta07: I think we agreed that for now, the feature where a user can override either DB or table is not supported. We will observe the usage of this feature and based on that improvise coz the usecase can be that "Put all selected tables in one DB".

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ct/SyTableReplication.java
abhimanyugupta07: Same reason as `SyReplicaTable`.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ct/SyTableReplication.java
abhimanyugupta07: Same reason as `ReplicaTable`

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ct/SyTableReplications.java
abhimanyugupta07: Then why even specify this tag in the config ? If the tag is specified then it needs to be completed.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ct/SyTableReplication.java
abhimanyugupta07: Yeah. Thought about that but decided against. Now that you are saying, I have added that as well. Its named as `SySourceTable`.

shunting-yard-replicator/pom.xml
abhimanyugupta07: Javax ExpressionFactory games.

You get `Caused by: java.lang.NoSuchMethodError: javax.el.ExpressionFactory.newInstance()` error while running test otherwise.

shunting-yard-binary/src/main/resources/shunting-yard-minimal.yml.template
abhimanyugupta07: Its not used in a test. Just a reference for all supported config options. I need to write a proper README for Shunting Yard similar to what we have for CT. For now, It acts as a good spot from which you can copy-paste config.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ct/SyReplicaTable.java
patduin: I don't remember agreeing on that. it's trivial to make it optional

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ct/SyTableReplications.java
patduin: Ok this means you can't say:
```
table-replications:
```
Right ok, I thought it meant you have to specify at least one table-replication. I'm ok if we keep it also don't mind if it goes don't see the problem with above in the the yml but perhaps I'm still missing something

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/receiver/ContextFactory.java
massdosage: We might need to have a huddle to agree on this...

massdosage (changes requested):
Looking like we're getting close with this. Are we also going to add some documentation to the README as to how this new feature works?

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/receiver/ContextFactory.java
abhimanyugupta07: So, i have made the `db-name` and `table-name` optional for `replica-table` but an incomplete configuration eg. provide `source-table` but not provide anything for `replica-table` is not acceptable.

Also, `table-replications` by itself is optional as a whole.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ct/SyReplicaTable.java
abhimanyugupta07: This class is removed as I am reusing CT one's now.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ct/SyReplicaTable.java
abhimanyugupta07: Its done.

shunting-yard-replicator/pom.xml
```diff
@@ -40,6 +41,10 @@
       <version>${project.version}</version>
     </dependency>
 
+
+
+    <!-- -->
```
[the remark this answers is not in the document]
abhimanyugupta07: removed.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ct/SyTableReplication.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  public void setReplicaTable(ReplicaTable replicaTable) {
+    this.replicaTable = replicaTable;
+  }
+
+  public String getReplicaDatabaseName() {
+    SourceTable sourceTable = getSourceTable();
+    ReplicaTable replicaTable = getReplicaTable();
+    String databaseName = replicaTable.getDatabaseName() != null ? replicaTable.getDatabaseName()
```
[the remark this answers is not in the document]
abhimanyugupta07: Yes.

shunting-yard-replicator/pom.xml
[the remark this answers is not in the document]
massdosage: Still there! as well as the two blank lines above it!

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ShuntingYardTableReplicationsMap.java
```diff
@@ -0,0 +1,45 @@
...
+        String key = (String
+            .join(".", tableReplication.getSourceTable().getDatabaseName().toLowerCase(),
+                tableReplication.getSourceTable().getTableName().toLowerCase()));
+        tableReplicationsMap.put(key, tableReplication);
+      }
+    }
+  }
+
+  public ShuntingYardTableReplication getTableReplication(String dbName, String tableName) {
+    String key = String.join(".", dbName.toLowerCase(), tableName.toLowerCase());
```
massdosage: Put `String.join(".", dbName.toLowerCase(), tableName.toLowerCase())` into a `createKey` method? Then you can use it here and above? (I've seen this code in so many places now it probably needs to go into some utils class but for now a private method here will do)

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ShuntingYardTableReplicationsMap.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import com.hotels.shunting.yard.replicator.exec.conf.ct.ShuntingYardTableReplication;
+import com.hotels.shunting.yard.replicator.exec.conf.ct.ShuntingYardTableReplications;
+
+public class ShuntingYardTableReplicationsMap {
+  private final Map<String, ShuntingYardTableReplication> tableReplicationsMap = new HashMap<>();
+
+  public ShuntingYardTableReplicationsMap() {}
```
massdosage: Default constructor not needed, compiler will do this for you.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/event/MetaStoreEvent.java
```diff
@@ -185,6 +211,10 @@ public String getQualifiedTableName() {
     return String.join(".", getDatabaseName(), getTableName());
   }
 
+  public String getQualifiedReplicaTableName() {
+    return String.join(".", getReplicaDatabaseName(), getReplicaTableName());
```
massdosage: Ah, here it is again, perhaps it does justify a higher level `TableDatabaseNameJoiner` util class?

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/MessageReaderAdapter.java
```diff
@@ -59,8 +66,21 @@ public void close() throws IOException {
   }
 
   private MetaStoreEvent map(ListenerEvent listenerEvent) {
+
```
massdosage: blank line alert

massdosage (changes requested):
A few comments in this round, over to you!

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/receiver/ContextFactory.java
massdosage: OK, so if they don't specify `table-replications` they get the default behaviour which is that the db and table names from the source are used for the replica. If they *do* provide `table-replications` then we're saying they have to provide both the replica db *and* table name?

abhimanyugupta07: Yes for the first statement and No for the second.

eg.

Valid 
```
table-replications:
  - source-table:
      database-name: bdp
      table-name: etl_daily_unique_visitors
    replica-table:
      table-name: etl_daily_unique_visitors_1

```
Invalid
```
table-replications:
  - source-table:
      database-name: bdp
      table-name: etl_daily_unique_visitors

```

shunting-yard-replicator/pom.xml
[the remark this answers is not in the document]
abhimanyugupta07: oh, there was one more commented one which I removed earlier. I have removed this as well.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/messaging/MessageReaderAdapter.java
abhimanyugupta07: removed

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ShuntingYardTableReplicationsMap.java
abhimanyugupta07: I don't think, this happens in Java.

https://www.codebyamir.com/blog/java-essentials-the-default-constructor

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/event/MetaStoreEvent.java
abhimanyugupta07: Ok. I have created that joiner.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ShuntingYardTableReplicationsMap.java
abhimanyugupta07: done.

abhimanyugupta07:
@massdosage All suggestions are incorporated. For the README, I think I will need to create a comprehensive document just like we have for CT before I can add these details in that. So, I will do that as a separate task. 

This PR is ready for another look.

shunting-yard-replicator/pom.xml
```diff
@@ -75,6 +76,12 @@
     <dependency>
       <groupId>org.apache.hadoop</groupId>
       <artifactId>hadoop-common</artifactId>
+      <exclusions>
```
patduin: Maybe add: ```          <!-- Exclude add because it conflicts with org.glassfish.avax.el dependency    -->```

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ShuntingYardTableReplicationsMap.java
patduin: that links is when a super class is involved.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ct/SyReplicaTable.java
patduin: ok so I know we started out with reusing this but then the validation became an issue. I'm not sure we should use it now. Consider the following in CT we add a validation to one of the ReplicaSource fields now we bump the version in SY to use that. Everything compiles just fine. But we'll get runtime exception when running stuff. Not sure we if we will catch all of that in tests. thoughts?

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ShuntingYardTableReplicationsMap.java
abhimanyugupta07: If I remove this, I get compilation error in the class where I am using it.

The link above says: "If we do explicitly declare a constructor of any form, then this automatic insertion by the compiler won't occur. "

Not sure if I am doing something wrong here.

shunting-yard-replicator/pom.xml
abhimanyugupta07: done.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ct/SyReplicaTable.java
abhimanyugupta07: Hmm.. good point.. I am only using dbName and tableName fields here. So, I feel like those two fields will perhaps not get any extra validation checks on them coz they are optional in CT.

Its just that if we copy those classes, it all looks a little ugly coz I need to use CT's ReplicaTable and SourceTable classes to build CT Config (In ContextFactory). It then causes confusion in the project to have both of them. But I will go with what everyone agrees on.

patduin: doesn't matter what you use. if a field gets added with @notNull validation it will fail. I know this would mean a major CT version and we should notice, nonetheless everything will compile it will either be caught in a test (best case) or at runtime (worst case). We do have this problem when building the CT config as well though so maybe it's ok. Dunno :(

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ShuntingYardTableReplicationsMap.java
massdosage: Well, there's always a super class, even if just the implicit `Object`. I'm confused why this is needed, what's the compiler error?

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/util/TableDatabaseNameJoiner.java
```diff
@@ -0,0 +1,9 @@
+package com.hotels.shunting.yard.replicator.util;
+
+public class TableDatabaseNameJoiner {
```
massdosage: Since this is a util class give it a private constructor

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/exec/conf/ShuntingYardTableReplicationsMap.java
abhimanyugupta07: This is removed as it was being used on in the test. Both test and the default constructor are now removed.

shunting-yard-replicator/src/main/java/com/hotels/shunting/yard/replicator/util/TableDatabaseNameJoiner.java
abhimanyugupta07: Done.

massdosage (approved):

patduin (approved):

PR #35, opened 2019-04-12
README for Shunting Yard

abhimanyugupta07:
First attempt at an exhaustive README for Shunting Yard.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ You can obtain Shunting Yard from Maven Central:
 
 ![Shunting Yard system diagram.](shunting-yard.png "Selected tables getting replicated by Shunting Yard based on Hive events.")
 
-### Install
+## Install
 
 1. Download the version to use from [Maven Central](https://mvnrepository.com/artifact/com.hotels/shunting-yard-binary) and uncompress it in a directory of your choosing.
 
@@ -37,7 +37,7 @@ Note that the paths above are correct as of when this document was last updated 
 
 The examples below all demonstrate configuration using YAML and provide fragments covering the most common use cases that should be useful as a basis for building your own configuration. A full configuration reference is provided in the following sections.
 
-#### Configuring source, replica and SQS queue
+### Configuring source, replica and SQS queue
 
 The YAML fragment below shows some common options for setting up the base source (where data is coming from), replica (where data is going to) and the SQS queue to read hive events from.
 
@@ -57,7 +57,7 @@ The YAML fragment below shows some common options for setting up the base source
     table-replications:
       ...
 
-#### Selecting tables to monitor
+### Selecting tables to monitor
 
 The YAML fragment below shows how to select the tables to be monitored by Shunting Yard.
 
@@ -66,11 +66,11 @@ The YAML fragment below shows how to select the tables to be monitored by Shunti
         - test_database.test_table_1
         - test_database.test_table_2
 
-#### Specifying target database & table names
+### Specifying target database & table names
 
 The YAML fragments below shows some common options for specifying the target database and table name for the selected tables.
 
-##### Specify both target database and table name.
+#### Specify both target database and table name.
 
     table-replications:
       - source-table:
@@ -80,7 +80,7 @@ The YAML fragments below shows some common options for specifying the target dat
           database-name: replica_database
           table-name: test_table_1    
           
-##### Only Change the target database but the table name remains same as source
+#### Only Change the target database but the table name remains same as source
 
     table-replications:
       - source-table:
@@ -89,7 +89,7 @@ The YAML fragments below shows some common options for specifying the target dat
         replica-table:
           database-name: replica_database 
 
-##### Only Change the target table name but the database remains same as source
+#### Only Change the target table name but the database remains same as source
 
     table-replications:
       - source-table:
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ You can obtain Shunting Yard from Maven Central:
 
 ## Overview
 
-Shunting Yard is intended to be a constantly running service which listens to a queue for Hive events. These events are emitted from the Hive metastore based on the operations performed on the Hive tables. For instance, an `ADD_PARTITION_EVENT` is emitted from the Hive metastore when a new partition is added to a table. Similarly, a `CREATE_TABLE_EVENT` is emitted when a new table is created in the Hive metastore. We recommend using [Apiary Metastore Listener](https://github.com/ExpediaGroup/apiary-extensions/tree/master/apiary-metastore-listener) for getting these events from your Hive Metastore.
+Shunting Yard is intended to be a constantly running service which listens to a queue for Hive events. These events are emitted from the Hive Metastore based on the operations performed on the Hive tables. For instance, an `ADD_PARTITION_EVENT` is emitted from the Hive Metastore when a new partition is added to a table. Similarly, a `CREATE_TABLE_EVENT` is emitted when a new table is created in the Hive Metastore. We recommend using [Apiary Metastore Listener](https://github.com/ExpediaGroup/apiary-extensions/tree/master/apiary-metastore-listener) for getting these events from your Hive Metastore.
 
 Once Shunting Yard receives an event from the queue, it extracts the relevant information from it to build a YAML file which it then passes on to [Circus Train](https://github.com/HotelsDotCom/circus-train) to perform the replication. Shunting Yard also aggregates a series of events so that a minimum number of replications are performed via Circus Train.
 
@@ -141,9 +141,9 @@ The table below describes all the available configuration values for Shunting Ya
 |Property|Required|Description|
 |:----|:----:|:----|
 |`source-catalog.name`|Yes|A name for the source catalog for events and logging.|
-|`source-catalog.hive-metastore-uris`|No|Fully qualified URI of the source cluster's Hive metastore Thrift service.|
+|`source-catalog.hive-metastore-uris`|Yes|Fully qualified URI of the source cluster's Hive Metastore Thrift service.|
 |`replica-catalog.name`|Yes|A name for the replica catalog for events and logging.|
-|`replica-catalog.hive-metastore-uris`|Yes|Fully qualified URI of the replica cluster's Hive metastore Thrift service.|
+|`replica-catalog.hive-metastore-uris`|Yes|Fully qualified URI of the replica cluster's Hive Metastore Thrift service.|
 |`event-receiver.configuration-properties.com.hotels.shunting.yard.event.receiver.sqs.queue`|Yes|Fully qualified URI of the [AWS SQS](https://aws.amazon.com/sqs/) Queue to read the Hive events from.|
 |`event-receiver.configuration-properties.com.hotels.shunting.yard.event.receiver.sqs.wait.time.seconds`|No|Wait time in seconds for which the receiver will poll the SQS queue for a batch of messages. Default is 10 seconds. Read more about long polling with AWS SQS [here](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-long-polling.html).|
 |`source-table-filter.table-names`|No|A list of tables selected for Shunting Yard replication. Supported format: `database_1.table_1, database_2.table_2`. If these are not provided, Shunting Yard will not replicate any table.|
--- a/README.md
+++ b/README.md
@@ -18,8 +18,8 @@ You can obtain Shunting Yard from Maven Central:
 
 2. Download and install the latest version of [Circus Train](http://mvnrepository.com/artifact/com.hotels/circus-train/) and set the `CIRCUS_TRAIN_HOME` environment variable:
 
-     export CIRCUS_TRAIN_HOME=/home/hadoop/circus-train-<circus-train-version>
-
+           export CIRCUS_TRAIN_HOME=/home/hadoop/circus-train-<circus-train-version>
+           
 ## Usage
 To run Shunting Yard you just need to execute the `bin/replicator.sh` script in the installation directory and pass the configuration file: 
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Shunting Yard
 
-Shunting Yard reads serialized Hive MetaStore Events from a queue ([AWS SQS](https://aws.amazon.com/sqs/) is currently supported) and replicates the data between two data lakes. It does this by building a YAML file with the information provided in the event which is then passed to [Circus Train](https://github.com/HotelsDotCom/circus-train) to perform the replication.
+Shunting Yard reads serialized Hive Metastore Events from a queue ([AWS SQS](https://aws.amazon.com/sqs/) is currently supported) and replicates the data between two data lakes. It does this by building a YAML file with the information provided in the event which is then passed to [Circus Train](https://github.com/HotelsDotCom/circus-train) to perform the replication.
 
 ## Start using
 
@@ -59,7 +59,7 @@ export HCAT_LIB=/usr/lib/hive-hcatalog/share/hcatalog/
 export HIVE_LIB=/usr/lib/hive/lib/
 ```
 
-Note that the paths above are correct as of when this document was last updated but may differ across EMR versions. Refer to the [EMR release guide](http://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-release-components.html) for more up to date information if necessary.
+Note that the paths above are correct as of when this document was last updated but may differ across EMR versions. Refer to the [EMR release guide](http://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-release-components.html) for more up-to-date information if necessary.
 
 ## Configuring Shunting Yard
 
@@ -101,7 +101,7 @@ Shunting Yard by default will not replicate any tables unless they are selected 
 
 ### Specifying target database & table names
 
-Shunting Yard will by default replicate the data into the replica data lake with same replica database name and table name as the source. Sometimes a user might need to change the replica database name or table name or both. The YAML fragments below shows some common options for specifying the replica database and table name for the selected tables.
+Shunting Yard will by default replicate data into the replica data lake with the same replica database name and table name as the source. Sometimes a user might need to change the replica database name or table name or both. The YAML fragments below shows some common options for specifying the replica database and table name for the selected tables.
 
 #### Specify both replica database and table name
 
@@ -115,7 +115,7 @@ Shunting Yard will by default replicate the data into the replica data lake with
           
 #### Change only the replica database but the table name remains same as source
 
-In this case, the replica table name is not provided in the `table-replications` and hence, it will be same as source table name.
+In this case, the replica table name is not provided in the `table-replications` and therefore, it will be same as source table name.
 
     table-replications:
       - source-table:
@@ -126,7 +126,7 @@ In this case, the replica table name is not provided in the `table-replications`
 
 #### Change only the replica table name but the database remains same as source
 
-In this case, the replica database name is not provided in the `table-replications` and hence, it will be same as source database name.
+In this case, the replica database name is not provided in the `table-replications` and therefore, it will be same as source database name.
 
     table-replications:
       - source-table:
--- a/README.md
+++ b/README.md
@@ -24,19 +24,19 @@ Once Shunting Yard receives an event from the queue, it extracts the relevant in
 
 [Download the TGZ](https://repository.sonatype.org/service/local/artifact/maven/redirect?r=central-proxy&g=com.hotels&a=shunting-yard&p=tgz&v=RELEASE&c=bin) from Maven central and then uncompress the file by executing:
 
-    tar -xzf shunting-yard-binary-<version>-bin.tgz
+        tar -xzf shunting-yard-binary-<version>-bin.tgz
     
-Although it's not necessary, we recommend exporting the environment variable SHUNTING_YARD_HOME by setting its value to wherever you extracted it to:
+Although it's not necessary, we recommend exporting the environment variable `SHUNTING_YARD_HOME` by setting its value to wherever you extracted it to:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,61 +1,147 @@
 # Shunting Yard
 
-A Spring Boot app that reads serialized Hive MetaStore Events and builds a YAML file with the information provided in the event which is then passed to [Circus Train](https://github.com/HotelsDotCom/circus-train) to perform the replication.
+Shunting Yard reads serialized Hive MetaStore Events from a queue (currently supports [AWS SQS](https://aws.amazon.com/sqs/)) and replicates the data between two Data lakes. It does this by building a YAML file with the information provided in the event which is then passed to [Circus Train](https://github.com/HotelsDotCom/circus-train) to perform the replication.
```
AnanaMJ: "between two Data lakes": does "data" have to be Uppercase?

README.md
```diff
@@ -1,61 +1,147 @@
...
 
-On the source cluster, copy the file respective emitter JAR to _/usr/lib/hive/lib/_ and configure _/etc/hive/conf/hive-site.xml_ with the properties required by the listener to talk to the messaging infrastructure:
+         export HCAT_LIB=/usr/lib/hive-hcatalog/share/hcatalog/
+         export HIVE_LIB=/usr/lib/hive/lib/
 
-#### SQS
-    hive.metastore.event.listeners = com.hotels.shunting.yard.event.emitter.sqs.listener.SqsMetaStoreEventListener
-    com.hotels.shunting.yard.event.emitter.sqs.queue = https://sqs.<region>.amazonaws.com/<account-id>/<topic-name>-queue.fifo
-    com.hotels.shunting.yard.event.emitter.sqs.group.id = <group-id>
+Note that the paths above are correct as of when this document was last updated but may differ across EMR versions, refer to the [EMR release guide](http://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-release-components.html) for more up to date information if necessary.
```
AnanaMJ: There should be a full stop instead of comma, as there seem to be two separate sentences in the same sentence.

README.md
```diff
@@ -1,61 +1,147 @@
...
+
+    table-replications:
+      - source-table:
+          database-name: source_database
+          table-name: test_table
+        replica-table:
+          database-name: replica_database
+          table-name: test_table_1    
+          
+#### Only Change the target database but the table name remains same as source
```
AnanaMJ: "change" with lowercase "c"

AnanaMJ: And maybe it would be good to rephrase this as "Change only the target database while keeping the same source table name" to keep the same tone of the writing (as an instruction)

README.md
```diff
@@ -1,61 +1,147 @@
...
+
+    table-replications:
+      - source-table:
+          database-name: source_database
+          table-name: test_table
+        replica-table:
+          database-name: replica_database 
+
+#### Only Change the target table name but the database remains same as source
```
AnanaMJ: Same thing here, but the database and table switched around :)

README.md
```diff
@@ -1,61 +1,147 @@
...
+### Shunting Yard configuration reference
+The table below describes all the available configuration values for Shunting Yard.
+
+|Property|Required|Description|
+|:----|:----:|:----|
+|`source-catalog.name`|Yes|A name for the source catalog for events and logging.|
+|`source-catalog.hive-metastore-uris`|No|Fully qualified URI of the source cluster's Hive metastore Thrift service.|
+|`replica-catalog.name`|Yes|A name for the replica catalog for events and logging.|
+|`replica-catalog.hive-metastore-uris`|Yes|Fully qualified URI of the replica cluster's Hive metastore Thrift service.|
+|`event-receiver.configuration-properties.com.hotels.shunting.yard.event.receiver.sqs.queue`|Yes|Fully qualified URI of the [AWS SQS](https://aws.amazon.com/sqs/) Queue to read the hive events from.|
```
AnanaMJ: Uppercase H on "hive"

README.md
```diff
@@ -1,61 +1,147 @@
...
+|`event-receiver.configuration-properties.com.hotels.shunting.yard.event.receiver.sqs.wait.time.seconds`|No|Wait time in seconds for which the receiver will poll the SQS queue for a batch of messages. Default is 10 seconds. Read more about long polling with AWS SQS [here](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-long-polling.html)|
```
AnanaMJ: Full stop at the end 🙂

README.md
```diff
@@ -1,61 +1,147 @@
...
+|`source-table-filter.table-names`|No|A list of tables selected for Shunting Yard replication. Supported Format:`database_1.table_1, database_2.table_2`|
```
AnanaMJ: space after colon:
"format: `database_1.table_1, database_2.table_2`"

and lowercase f on "Format"

README.md
```diff
@@ -1,61 +1,147 @@
...
+|`table-replications[n].source-table.database-name`|No|The name of the database in which the table you wish to replicate is located.|
+|`table-replications[n].source-table.table-name`|No|The name of the table which you wish to replicate.|
+|`table-replications[n].replica-table.database-name`|No|The name of the destination database in which to replicate the table. Defaults to source database name.|
+|`table-replications[n].replica-table.table-name`|No|The name of the table at the destination. Defaults to source table name.|
+
+### Configuring Graphite Metrics
```
AnanaMJ: I think "Metrics" should be with lowercase m. "System architecture" has 2 hashes (compared to 3 here) and doesn't have uppercase for each letter in the subtitle, so this shouldn't either.

abhimanyugupta07:
> I definitely wasn't looking for a missing full stop on purpose :)

I think you were 🙈

README.md
abhimanyugupta07: done.

README.md
abhimanyugupta07: done.

README.md
abhimanyugupta07: done.

README.md
abhimanyugupta07: done.

README.md
abhimanyugupta07: done.

README.md
abhimanyugupta07: done.

README.md
abhimanyugupta07: done.

README.md
abhimanyugupta07: done.

README.md
AnanaMJ: what about the second half?

README.md
AnanaMJ: There's still no space after colon

README.md
AnanaMJ: Graphite can stay with uppercase, because it's the name of the tool, right?

README.md
```diff
@@ -1,6 +1,6 @@
 # Shunting Yard
 
-Shunting Yard reads serialized Hive MetaStore Events from a queue (currently supports [AWS SQS](https://aws.amazon.com/sqs/)) and replicates the data between two Data lakes. It does this by building a YAML file with the information provided in the event which is then passed to [Circus Train](https://github.com/HotelsDotCom/circus-train) to perform the replication.
+Shunting Yard reads serialized Hive MetaStore Events from a queue (currently supports [AWS SQS](https://aws.amazon.com/sqs/)) and replicates the data between two data lakes. It does this by building a YAML file with the information provided in the event which is then passed to [Circus Train](https://github.com/HotelsDotCom/circus-train) to perform the replication.
```
massdosage: currently supports [AWS SQS](https://aws.amazon.com/sqs/) -> [AWS SQS](https://aws.amazon.com/sqs/) is currently supported

README.md
```diff
@@ -1,61 +1,147 @@
...
 
 ## System architecture
 
 ![Shunting Yard system diagram.](shunting-yard.png "Selected tables getting replicated by Shunting Yard based on Hive events.")
 
-## Start using
+## Install
 
-You can obtain Shunting Yard from Maven Central:
+1. Download the version to use from [Maven Central](https://mvnrepository.com/artifact/com.hotels/shunting-yard-binary) and uncompress it in a directory of your choosing.
```
massdosage: It's a bit unclear here which file one should actually download, for a new user we really need to spell it out for them. I know this is what it says in the Circus Train docs but that could also be improved (once we get the wording sorted out here let's do the same for Circus Train). I think Waggle Dance is a better example of this: https://github.com/HotelsDotCom/waggle-dance/#install

README.md
```diff
@@ -1,61 +1,147 @@
...
 
-[![Maven Central](https://maven-badges.herokuapp.com/maven-central/com.hotels/shunting-yard/badge.svg?subject=com.hotels:shunting-yard)](https://maven-badges.herokuapp.com/maven-central/com.hotels/shunting-yard) [![Build Status](https://travis-ci.org/HotelsDotCom/shunting-yard.svg?branch=master)](https://travis-ci.org/HotelsDotCom/shunting-yard) [![Coverage Status](https://coveralls.io/repos/github/HotelsDotCom/shunting-yard/badge.svg?branch=master)](https://coveralls.io/github/HotelsDotCom/shunting-yard?branch=master) ![GitHub license](https://img.shields.io/github/license/HotelsDotCom/shunting-yard.svg)
+2. Download and install the latest version of [Circus Train](http://mvnrepository.com/artifact/com.hotels/circus-train/) and set the `CIRCUS_TRAIN_HOME` environment variable:
```
massdosage: Once we sort out the "Install" section for Circus Train let's rather link to the section in its README instead of duplicating links to Maven repos etc. here.

README.md
```diff
@@ -1,61 +1,147 @@
...
+
+         export CIRCUS_TRAIN_HOME=/home/hadoop/circus-train-<circus-train-version>
```
massdosage: Should we use `/opt` as a base instead of `/home/hadoop` in the examples?

README.md
```diff
@@ -1,61 +1,147 @@
...
+## Configuring Shunting Yard
 
-    sudo restart hive-hcatalog-server
+The examples below all demonstrate configuration using YAML and provide fragments covering the most common use cases that should be useful as a basis for building your own configuration. A full configuration reference is provided in the following sections.
 
-### How to install the receiver - TARGET
+### Configuring source, replica and SQS queue
 
-On the target cluster, download and install the latests version of Circus Train and set the `CIRCUS_TRAIN_HOME` environment variable:
+The YAML fragment below shows some common options for setting up the base source (where data is coming from), replica (where data is going to) and the SQS queue to read hive events from.
```
massdosage: Hive

README.md
```diff
@@ -1,61 +1,147 @@
...
+You can obtain Shunting Yard from Maven Central:
+
+[![Maven Central](https://maven-badges.herokuapp.com/maven-central/com.hotels/shunting-yard/badge.svg?subject=com.hotels:shunting-yard)](https://maven-badges.herokuapp.com/maven-central/com.hotels/shunting-yard) [![Build Status](https://travis-ci.org/HotelsDotCom/shunting-yard.svg?branch=master)](https://travis-ci.org/HotelsDotCom/shunting-yard) [![Coverage Status](https://coveralls.io/repos/github/HotelsDotCom/shunting-yard/badge.svg?branch=master)](https://coveralls.io/github/HotelsDotCom/shunting-yard?branch=master) ![GitHub license](https://img.shields.io/github/license/HotelsDotCom/shunting-yard.svg)
 
 ## System architecture
 
 ![Shunting Yard system diagram.](shunting-yard.png "Selected tables getting replicated by Shunting Yard based on Hive events.")
 
-## Start using
+## Install
```
massdosage: I'd recommend an "Overview" section where you explain in a bit more detail what it does and how it works. It can either be a paragraph or two which basically describes the diagram above in words and/or a list of typical steps that happen (a bit like what @AnanaMJ did for describing Housekeeping as a service). I think it would be good for a user to know this before they move on to installing it and configuring and running it.

README.md
```diff
@@ -1,61 +1,147 @@
...
+The examples below all demonstrate configuration using YAML and provide fragments covering the most common use cases that should be useful as a basis for building your own configuration. A full configuration reference is provided in the following sections.
 
-### How to install the receiver - TARGET
+### Configuring source, replica and SQS queue
 
-On the target cluster, download and install the latests version of Circus Train and set the `CIRCUS_TRAIN_HOME` environment variable:
+The YAML fragment below shows some common options for setting up the base source (where data is coming from), replica (where data is going to) and the SQS queue to read hive events from.
 
-    export CIRCUS_TRAIN_HOME=/home/hadoop/circus-train-<circus-train-version>
+    source-catalog:
```
massdosage: Could we have the absolute bare minimum yaml file that would still actually work here? I think it's pretty much what you have but without the `source-table-filter` and `table-replications` sections? That would then replicate everything right? If this is possible that might be a good starting point (with a couple of sentences explaining this). Then the sections below could state that one can add these extra elements and explain what they do?

README.md
```diff
@@ -1,61 +1,147 @@
...
+        - ...
+    table-replications:
+      ...
 
-Export the location of the Hive libraries:
+### Selecting tables to monitor
 
-    export HIVE_LIB=/usr/lib/hive/lib/
-    export HCAT_LIB=/usr/lib/hive-hcatalog/share/hcatalog/
+The YAML fragment below shows how to select the tables to be monitored by Shunting Yard.
```
massdosage: Spell it out a bit more, something like "Shunting Yard by default will replicate all tables in a source Hive Metastore but the user can choose to limit this to a subset of tables by using a source table filter. For example to only replicate tables names `test_table_1` and `test_table_2` you would configure a `source-table-filter` like so:"

README.md
```diff
@@ -1,61 +1,147 @@
...
+    source-table-filter:
+      table-names:
+        - test_database.test_table_1
+        - test_database.test_table_2
 
-    export SHUNTING_YARD_HOME=/home/hadoop/shuting-yard-<version>
+### Specifying target database & table names
 
-Create a `.yml` file under `$SHUNTING_YARD_HOME/conf/` with the correct settings to talk to the messaging infrastructure. You can find guidelines in the file _shunting-yard-minimal.yml.template_. Once this is done you are all set up to run the service:
+The YAML fragments below shows some common options for specifying the target database and table name for the selected tables.
```
massdosage: Similar to above, spell it out what the feature is and then what exactly the example yaml fragment is doing.

README.md
```diff
@@ -1,61 +1,147 @@
...
+    table-replications:
+      - source-table:
+          database-name: source_database
+          table-name: test_table
+        replica-table:
+          database-name: replica_database
+          table-name: test_table_1    
+          
+#### Change only the target database but the table name remains same as source
+
```
massdosage: etc.

README.md
```diff
@@ -1,61 +1,147 @@
...
+    table-replications:
+      - source-table:
+          database-name: source_database
+          table-name: test_table
+        replica-table:
+          database-name: replica_database 
+
+#### Change only the target table name but the database remains same as source
+
```
massdosage: and here. For this and the one above you probably just need to describe what's going on in the yaml as the header already explains the feature.

README.md
```diff
@@ -1,61 +1,147 @@
...
+        replica-table:
+          table-name: test_table_1
+
+### Shunting Yard configuration reference
+The table below describes all the available configuration values for Shunting Yard.
+
+|Property|Required|Description|
+|:----|:----:|:----|
+|`source-catalog.name`|Yes|A name for the source catalog for events and logging.|
+|`source-catalog.hive-metastore-uris`|No|Fully qualified URI of the source cluster's Hive metastore Thrift service.|
```
massdosage: What happens if you don't provide this? We should probably explain...

README.md
```diff
@@ -1,61 +1,147 @@
...
+|`replica-catalog.name`|Yes|A name for the replica catalog for events and logging.|
+|`replica-catalog.hive-metastore-uris`|Yes|Fully qualified URI of the replica cluster's Hive metastore Thrift service.|
+|`event-receiver.configuration-properties.com.hotels.shunting.yard.event.receiver.sqs.queue`|Yes|Fully qualified URI of the [AWS SQS](https://aws.amazon.com/sqs/) Queue to read the Hive events from.|
```
massdosage: Somewhere we should explain that these are actually events coming from Apiary right? Perhaps in the overview right at the top where things are covered in more depth?

README.md
```diff
@@ -1,61 +1,147 @@
...
+|`event-receiver.configuration-properties.com.hotels.shunting.yard.event.receiver.sqs.wait.time.seconds`|No|Wait time in seconds for which the receiver will poll the SQS queue for a batch of messages. Default is 10 seconds. Read more about long polling with AWS SQS [here](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-long-polling.html).|
+|`source-table-filter.table-names`|No|A list of tables selected for Shunting Yard replication. Supported format:`database_1.table_1, database_2.table_2`|
```
massdosage: For all the values which are optional we should explain what the behaviour is if they're not set to anything.

README.md
```diff
@@ -1,61 +1,147 @@
...
+|`table-replications[n].source-table.database-name`|No|The name of the database in which the table you wish to replicate is located.|
+|`table-replications[n].source-table.table-name`|No|The name of the table which you wish to replicate.|
+|`table-replications[n].replica-table.database-name`|No|The name of the destination database in which to replicate the table. Defaults to source database name.|
```
massdosage: Yes, here and the next one we explain what happens if not set.

README.md
```diff
@@ -1,61 +1,147 @@
...
+|`table-replications[n].replica-table.table-name`|No|The name of the table at the destination. Defaults to source table name.|
+
+### Configuring Graphite metrics
 
 Graphite configurations can be passed to Shunting Yard using an optional `--ct-config` argument which takes a YAML file and passes it directly to internal Circus Train instance. Refer to the [Circus Train README](https://github.com/HotelsDotCom/circus-train#graphite) for more details.
```
massdosage: Make clear this is a *different* YAML configuration file to the one described above.

README.md
abhimanyugupta07: Yeah

README.md
AnanaMJ: We talked about this and I'm ok with what's in the document now :)

README.md
massdosage: This ^

abhimanyugupta07: oh. That is mandatory. Changed now.

README.md
abhimanyugupta07: It is there now :)

README.md
```diff
@@ -1,61 +1,191 @@
 # Shunting Yard
 
-A Spring Boot app that reads serialized Hive MetaStore Events and builds a YAML file with the information provided in the event which is then passed to [Circus Train](https://github.com/HotelsDotCom/circus-train) to perform the replication.
+Shunting Yard reads serialized Hive MetaStore Events from a queue ([AWS SQS](https://aws.amazon.com/sqs/) is currently supported) and replicates the data between two data lakes. It does this by building a YAML file with the information provided in the event which is then passed to [Circus Train](https://github.com/HotelsDotCom/circus-train) to perform the replication.
```
AnanaMJ: MetaStore &rightarrow; Metastore

README.md
```diff
@@ -1,61 +1,191 @@
...
+If you are planning to run Shunting Yard on EMR you will need to set up the EMR classpath by exporting the following environment variables before calling the `bin/replicator.sh` script:
+
+```
+export HCAT_LIB=/usr/lib/hive-hcatalog/share/hcatalog/
+export HIVE_LIB=/usr/lib/hive/lib/
+```
 
-    export HIVE_LIB=/usr/lib/hive/lib/
-    export HCAT_LIB=/usr/lib/hive-hcatalog/share/hcatalog/
+Note that the paths above are correct as of when this document was last updated but may differ across EMR versions. Refer to the [EMR release guide](http://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-release-components.html) for more up to date information if necessary.
```
AnanaMJ: up to date information &rightarrow; up-to-date information

Up to date is without hyphens when it's used after the noun it describes: "The information is up to date."
When it's before the noun, it's with hyphens: "This is an up-to-date information."
source: https://dictionary.cambridge.org/dictionary/english/up-to-date

README.md
```diff
@@ -1,61 +1,191 @@
...
-Graphite configurations can be passed to Shunting Yard using an optional `--ct-config` argument which takes a YAML file and passes it directly to internal Circus Train instance. Refer to the [Circus Train README](https://github.com/HotelsDotCom/circus-train#graphite) for more details.
+    source-table-filter:
+      table-names:
+        - test_database.test_table_1
+        - test_database.test_table_2
 
-#### Sample ct-config.yml:
+### Specifying target database & table names
+
+Shunting Yard will by default replicate the data into the replica data lake with same replica database name and table name as the source. Sometimes a user might need to change the replica database name or table name or both. The YAML fragments below shows some common options for specifying the replica database and table name for the selected tables.
```
AnanaMJ: with same replica database name and table name &rightarrow; with **the** same replica database name and table name

README.md
```diff
@@ -1,61 +1,191 @@
...
+      - source-table:
+          database-name: source_database
+          table-name: test_table
+        replica-table:
+          database-name: replica_database
+          table-name: test_table_1    
+          
+#### Change only the replica database but the table name remains same as source
+
+In this case, the replica table name is not provided in the `table-replications` and hence, it will be same as source table name.
```
AnanaMJ: and hence &rightarrow; ~and~ hence
It's like saying "and therefore" or "and so"

README.md
AnanaMJ: If you want to not have that many "the", you can have 
"Shunting Yard will by default replicate data into a replica data lake with the same replica database name and table name as the source."

README.md
abhimanyugupta07: .... closes laptop and goes home ;)

PR #38, opened 2019-06-12
Excluded `org.pentaho.pentaho-aggdesigner-algorithm` dependency

abhimanyugupta07:
This dependency isn't available in Maven Central and means you can't build the project if you don't have a settings.xml file (i.e. Maven defaults).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) a
 ### Fixed
 * `FilteringMessageReader` now deletes filtered messages to mimic previous behaviour of delete on read.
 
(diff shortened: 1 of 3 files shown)


abhimanyugupta07:
> This looks fine but you're going to run into similar issues with `javax.jms:jms` (I just tried it). Try build this project from scratch with an empty .m2 cache and no `settings.xml` (easiest way to do this is probably to temporarily rename .m2 to something else and run a build locally and see what else you need to exclude. Hopefully it's just that jms library.

As we discussed on slack, I am not getting any errors with jms dependency on my local. We have tested this branch on jenkins and we are not getting any errors there too. So, we are going to go ahead without adding jms as exclusion and see if it gets released successfully.

PR #41, opened 2019-06-28
Upgraded `hotels-oss-parent` to 4.1.0

massdosage:
This was one of the projects I was testing the latest version of the parent pom out on. From what I can tell everything works fine, we'll need to do a release to test the removal of the GPG plugin but I don't think it's needed here any more.

Once we have a release version of the eg-oss-parent we'll move this over to use that but that's for a different PR ;)


massdosage:
> Should we mention in the Readme that a profile was removed?
> 
> In any case, looks good +1

The profile is still there, it's just in the parent (https://github.com/HotelsDotCom/hotels-oss-parent/blob/master/pom.xml#L472) instead of being overridden here.

PR #45, opened 2019-10-18
Support table parameters

max-jacobs:
Making a small update for SY to support the latest Circus Train and to allow skipping CT's housekeeping process.

Main change is in `CircusTrainRunner`

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,23 @@ In this case, the replica database name is not provided in the `table-replicatio
         replica-table:
           table-name: test_table_1
 
+### Orphaned data strategy
+
+Shunting Yard will invoke Circus Train's default configuration which is to run the Housekeeping process to cleanup dereferenced snapshots. This means that Housekeeping configuration must be provided.
+
+To override this, add the configuration parameter `orphaned-data-strategy`:
+
+    table-replications:
+      - source-table:
+          database-name: source_database
+          table-name: test_table
+        replica-table:
+          database-name: replica_database
+          table-name: test_table_1 
+    orphaned-data-strategy: NONE 
+ 
+This will ensure that Shunting Yard only starts up Circus Train's replication module, and that no paths are added to a Housekeeping database.
+
 ### Shunting Yard configuration reference
 The table below describes all the available configuration values for Shunting Yard.
 
@@ -147,6 +164,7 @@ The table below describes all the available configuration values for Shunting Ya
 |`sqs.queue`|Yes|Fully qualified URI of the [AWS SQS](https://aws.amazon.com/sqs/) Queue to read the Hive events from.|
 |`sqs.wait.time.seconds`|No|Wait time in seconds for which the receiver will poll the SQS queue for a batch of messages. Default is 10 seconds. Read more about long polling with AWS SQS [here](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-long-polling.html).|
 |`source-table-filter.table-names`|No|A list of tables selected for Shunting Yard replication. Supported format: `database_1.table_1, database_2.table_2`. If these are not provided, Shunting Yard will not replicate any tables.|
+|`orphaned-data-strategy`|No|Orphaned data strategy to use for replications. Possible values: `NONE` and `HOUSEKEEPING`. Default is `HOUSEKEEPING`.|
 |`table-replications[n].source-table.database-name`|No|The name of the database in which the table you wish to replicate is located. `table-replications` section is optional and if it is not provided, Shunting Yard will use the database name and table name from the source for the replica.|
 |`table-replications[n].source-table.table-name`|No|The name of the table which you wish to replicate.|
 |`table-replications[n].replica-table.database-name`|No|The name of the destination database in which to replicate the table. Defaults to `source-table.database-name`|
@@ -162,6 +180,18 @@ Graphite configuration can be passed to Shunting Yard using an optional `--ct-co
       host: graphite-host:2003
       namespace: com.company.shuntingyard
       prefix: dev
+
+### Configuring table transformations
+
+Circus Train can transform the metadata of replica tables by adding table parameters during the replication. Please see [Metadata transformations](https://github.com/HotelsDotCom/circus-train#metadata-transformations) in the Circus Train docs for more detailed instructions.
+
+#### Sample ct-config.yml with table transformations:
+
+    transform-options:
+      table-properties:
+        my-custom-property: my-custom-value
+        my-custom-property2: my-custom-value2
+
 
 ### Housekeeping
 
--- a/README.md
+++ b/README.md
@@ -181,18 +181,6 @@ Graphite configuration can be passed to Shunting Yard using an optional `--ct-co
       namespace: com.company.shuntingyard
       prefix: dev
 
-### Configuring table transformations
-
-Circus Train can transform the metadata of replica tables by adding table parameters during the replication. Please see [Metadata transformations](https://github.com/HotelsDotCom/circus-train#metadata-transformations) in the Circus Train docs for more detailed instructions.
-
-#### Sample ct-config.yml with table transformations:
-
-    transform-options:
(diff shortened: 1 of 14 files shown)


shunting-yard-replicator/src/main/java/com/expediagroup/shuntingyard/replicator/exec/conf/OrphanedDataStrategyConfig.java
```diff
@@ -0,0 +1,35 @@
...
+package com.expediagroup.shuntingyard.replicator.exec.conf;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+import com.hotels.bdp.circustrain.api.conf.OrphanedDataStrategy;
+
+@Configuration
+@ConfigurationProperties
+public class OrphanedDataStrategyConfig {
```
massdosage: OK, so we need a suffix on this one since it's not the actual `OrphanedDataStrategy`. I see we have one other class in this package called `EventReceiverConfiguration` - should we follow the naming convention and change `Config`->`Configuration`?

shunting-yard-replicator/src/main/java/com/expediagroup/shuntingyard/replicator/exec/conf/OrphanedDataStrategyConfig.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package com.expediagroup.shuntingyard.replicator.exec.conf;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+import com.hotels.bdp.circustrain.api.conf.OrphanedDataStrategy;
+
+@Configuration
+@ConfigurationProperties
```
massdosage: Doesn't this need a prefix "orphaned-data-strategy" to get bound to the value from the YAML?

README.md
```diff
@@ -147,6 +164,7 @@ The table below describes all the available configuration values for Shunting Ya
 |`sqs.queue`|Yes|Fully qualified URI of the [AWS SQS](https://aws.amazon.com/sqs/) Queue to read the Hive events from.|
 |`sqs.wait.time.seconds`|No|Wait time in seconds for which the receiver will poll the SQS queue for a batch of messages. Default is 10 seconds. Read more about long polling with AWS SQS [here](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-long-polling.html).|
 |`source-table-filter.table-names`|No|A list of tables selected for Shunting Yard replication. Supported format: `database_1.table_1, database_2.table_2`. If these are not provided, Shunting Yard will not replicate any tables.|
+|`orphaned-data-strategy`|No|Orphaned data strategy to use for replications. Possible values: `NONE` and `HOUSEKEEPING`. Default is `HOUSEKEEPING`.|
```
massdosage: I'm missing where this default happens in the code? I see you've added the ability to set an orphaned data strategy but is there a test (maybe an existing one so it doesn't show up in the PR) that shows that if you don't set this then the value is `HOUSEKEEPING`?

shunting-yard-replicator/src/main/java/com/expediagroup/shuntingyard/replicator/exec/conf/OrphanedDataStrategyConfig.java
max-jacobs: Yes good point, one day I'll name these config classes appropriately...

shunting-yard-replicator/src/main/java/com/expediagroup/shuntingyard/replicator/exec/conf/OrphanedDataStrategyConfig.java
max-jacobs: I don't _think_ so because in this case there are no "sub" properties. Spring is clever enough to know to map the yaml property "orphaned-data-strategy" to the field in here called "orphanedDataStrategy". I've tested this out on EMR doing an actual replication and it works. Shame there's no integration tests in SY like in CT to test out different config settings.

README.md
max-jacobs: this happens in `OrphanedDataStrategyConfig` line 26 the default is if no yaml configuration overrides this initial setting. 

So there's no test for this because i guess it would require an integration test of sorts in order to test the mapping from yaml to bean, and we don't have these in SY at the moment. should I make it part of this PR to add these?

shunting-yard-replicator/src/test/java/com/expediagroup/shuntingyard/replicator/exec/external/CircusTrainConfigTest.java
```diff
@@ -96,6 +96,7 @@ public void withMultiValuePartitionFilter() {
     assertThat(replication.getReplicaTable().getDatabaseName()).isEqualTo("replicaDatabaseName");
     assertThat(replication.getReplicaTable().getTableName()).isEqualTo("replicaTableName");
     assertThat(replication.getReplicaTable().getTableLocation()).isEqualTo("replicaTableLocation");
+    assertThat(replication.getOrphanedDataStrategy()).isEqualTo(OrphanedDataStrategy.HOUSEKEEPING);
```
abhimanyugupta07: Will be good to add a test when `OrphanedDataStrategy` is null.

The `CircusTrainConfig` doesn't do a null check on this argument at the moment.

shunting-yard-replicator/src/main/java/com/expediagroup/shuntingyard/replicator/exec/conf/OrphanedDataStrategyConfig.java
patduin: I think can be separate we already have an issue: https://github.com/ExpediaGroup/shunting-yard/issues/5

README.md
```diff
@@ -135,6 +135,23 @@ In this case, the replica database name is not provided in the `table-replicatio
...
+    table-replications:
+      - source-table:
+          database-name: source_database
+          table-name: test_table
+        replica-table:
+          database-name: replica_database
+          table-name: test_table_1 
+    orphaned-data-strategy: NONE 
+ 
+This will ensure that Shunting Yard only starts up Circus Train's replication module, and that no paths are added to a Housekeeping database.
```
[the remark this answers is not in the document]
max-jacobs: Yes may be good to provide some context, but Beekeeper isn't a "strategy" that's available as an options - only "none" and "housekeeping".  Something like this:

> This is a necessary step if you want to make use of an external housekeeping mechanism, for example [Beekeeper](https://github.com/ExpediaGroup/beekeeper).

?

README.md
```diff
@@ -163,6 +181,18 @@ Graphite configuration can be passed to Shunting Yard using an optional `--ct-co
       namespace: com.company.shuntingyard
       prefix: dev
 
+### Configuring table transformations
```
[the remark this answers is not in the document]
max-jacobs: Will update this

README.md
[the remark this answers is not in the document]
abhimanyugupta07: Sounds good. I have updated my suggestion.

shunting-yard-replicator/src/test/java/com/expediagroup/shuntingyard/replicator/exec/external/CircusTrainConfigTest.java
max-jacobs: Good point, will add this in, thanks

README.md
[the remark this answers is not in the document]
max-jacobs: Needs a comma after "data" not a full stop.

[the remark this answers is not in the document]
max-jacobs: I've managed to update and commit, thanks!

max-jacobs:
This now uses the latest Circus Train version, also tested this on EMR and it works, so going to merge.

PR #47, opened 2020-01-15
SY to perform FULL replication on insert overwrite events

abhimanyugupta07:
To achieve this, SY will need to make a callback to Source Hive Metastore to determine if a table is partitioned or not as we don't have that information in ALTER_TABLE event.

ALTER_TABLE event: https://github.com/ExpediaGroup/apiary-extensions/tree/master/apiary-metastore-events/sns-metastore-events/apiary-metastore-listener#alter-table

Other change is that for ALTER_PARTITION event, it will always be a FULL_UPDATE.

### DIFF
--- a/shunting-yard-replicator/src/main/java/com/expediagroup/shuntingyard/replicator/exec/context/CommonBeans.java
+++ b/shunting-yard-replicator/src/main/java/com/expediagroup/shuntingyard/replicator/exec/context/CommonBeans.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2016-2019 Expedia, Inc.
+ * Copyright (C) 2016-2020 Expedia, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -37,6 +37,7 @@ import org.springframework.core.Ordered;
 import org.springframework.core.annotation.Order;
 
 import com.google.common.base.Supplier;
+import com.google.common.collect.Lists;
 
 import com.expedia.apiary.extensions.receiver.common.messaging.MessageReader;
 
@@ -110,6 +111,16 @@ public class CommonBeans {
     return hiveConf;
   }
 
+  @Bean
+  HiveConf sourceHiveConf(Configuration baseConfiguration, SourceCatalog sourceCatalog) {
+    Map<String, String> properties = new HashMap<>();
+    if (sourceCatalog.getHiveMetastoreUris() != null) {
+      properties.put(ConfVars.METASTOREURIS.varname, sourceCatalog.getHiveMetastoreUris());
+    }
+    HiveConf hiveConf = new HiveConfFactory(Lists.newArrayList(), properties).newInstance();
+    return hiveConf;
+  }
+
   private void putConfigurationProperties(Map<String, String> configurationProperties, Map<String, String> properties) {
     if (configurationProperties != null) {
       properties.putAll(configurationProperties);
@@ -129,13 +140,20 @@ public class CommonBeans {
   }
 
   @Bean
+  Supplier<CloseableMetaStoreClient> sourceMetaStoreClientSupplier(
+      HiveConf sourceHiveConf,
+      MetaStoreClientFactory sourceMetaStoreClientFactory) {
+    return new DefaultMetaStoreClientSupplier(sourceHiveConf, sourceMetaStoreClientFactory);
+  }
+
+  @Bean
   ReplicationMetaStoreEventListener replicationMetaStoreEventListener(
       HiveConf replicaHiveConf,
       Supplier<CloseableMetaStoreClient> replicaMetaStoreClientSupplier,
       OrphanedDataStrategyConfiguration orphanedDataStrategyConfiguration) {
     CloseableMetaStoreClient metaStoreClient = replicaMetaStoreClientSupplier.get();
     ContextFactory contextFactory = new ContextFactory(replicaHiveConf, metaStoreClient, new Marshaller(),
-      orphanedDataStrategyConfiguration.getOrphanedDataStrategy());
+        orphanedDataStrategyConfiguration.getOrphanedDataStrategy());
     return new CircusTrainReplicationMetaStoreEventListener(metaStoreClient, contextFactory, new CircusTrainRunner());
   }
 
@@ -152,6 +170,7 @@ public class CommonBeans {
   @Bean
   MessageReaderAdapter messageReaderAdapter(
       HiveConf replicaHiveConf,
+      Supplier<CloseableMetaStoreClient> sourceMetaStoreClientSupplier,
       EventReceiverConfiguration messageReaderConfig,
       SourceCatalog sourceCatalog,
       TableSelector tableSelector,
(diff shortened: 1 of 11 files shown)


shunting-yard-replicator/src/main/java/com/expediagroup/shuntingyard/replicator/exec/messaging/MessageReaderAdapter.java
```diff
@@ -138,4 +147,18 @@ private MetaStoreEvent map(ListenerEvent listenerEvent) {
     return builder.build();
   }
 
+  private boolean isPartitionedTable(String dbName, String tableName) {
+    Table sourceTable = null;
+    try {
+      sourceTable = sourceMetastoreClient.getTable(dbName, tableName);
+    } catch (TException e) {
```
patduin: `return sourceTable.getPartitionKeysSize() > 0;`

remove everything after trycatch: line158

shunting-yard-replicator/src/main/java/com/expediagroup/shuntingyard/replicator/exec/context/CommonBeans.java
```diff
@@ -110,6 +111,16 @@ HiveConf replicaHiveConf(
     return hiveConf;
   }
 
+  @Bean
+  HiveConf sourceHiveConf(Configuration baseConfiguration, SourceCatalog sourceCatalog) {
```
max-jacobs: Is it fine that there are now two `HiveConf` beans? I guess spring knows which one to inject by the variable/method name?

shunting-yard-replicator/src/main/java/com/expediagroup/shuntingyard/replicator/exec/messaging/MessageReaderAdapter.java
```diff
@@ -121,9 +128,11 @@ private MetaStoreEvent map(ListenerEvent listenerEvent) {
       break;
     case ALTER_TABLE:
       AlterTableEvent alterTable = (AlterTableEvent) listenerEvent;
-      if (alterTable.getTableLocation() != null) {
-        if (alterTable.getTableLocation().equals(alterTable.getOldTableLocation())) {
-          builder.replicationMode(ReplicationMode.METADATA_UPDATE);
+      if (isPartitionedTable(alterTable.getDbName(), alterTable.getTableName())) {
```
max-jacobs: Not a big deal but maybe can we unnest one of these ifs? Might read a bit better.

Could do a check before `if (!isPartitionedTable(...)) { break; }`

shunting-yard-replicator/src/main/java/com/expediagroup/shuntingyard/replicator/exec/messaging/MessageReaderAdapter.java
```diff
@@ -138,4 +147,18 @@ private MetaStoreEvent map(ListenerEvent listenerEvent) {
...
 
+  private boolean isPartitionedTable(String dbName, String tableName) {
+    Table sourceTable = null;
+    try {
+      sourceTable = sourceMetastoreClient.getTable(dbName, tableName);
+    } catch (TException e) {
+      throw new ShuntingYardException(String.format("Could not find table {}.{}", dbName, tableName), e);
+    }
+
+    if (CollectionUtils.isEmpty(sourceTable.getPartitionKeys())) {
```
max-jacobs: Also not a big deal but could be:

`return !CollectionUtils.isEmpty(sourceTable.getPartitionKeys())` right after line 153 in the try block.

shunting-yard-replicator/src/main/java/com/expediagroup/shuntingyard/replicator/exec/messaging/MessageReaderAdapter.java
```diff
@@ -69,7 +81,7 @@ public void close() throws IOException {
     }
   }
 
-  private MetaStoreEvent map(ListenerEvent listenerEvent) {
+  private MetaStoreEvent map(ListenerEvent listenerEvent) throws ShuntingYardException {
```
max-jacobs: Does there need to be a throws here?

shunting-yard-replicator/src/test/java/com/expediagroup/shuntingyard/replicator/exec/messaging/MessageReaderAdapterTest.java
```diff
@@ -359,12 +363,13 @@ public void metadataOnlySyncEventForTable() {
...
   @Test
-  public void metadataOnlySyncEventForTableWithNullLocation() {
+  public void insertOverwriteEventForUnpartitionedTable() throws MetaException, NoSuchObjectException, TException {
     Optional<MessageEvent> event = newMessageEvent(apiaryAlterTableEvent);
     when(messageReader.read()).thenReturn(event);
     configureMockedEvent(apiaryAlterTableEvent);
 
-    when(apiaryAlterTableEvent.getTableLocation()).thenReturn(null);
+    when(metaStoreClient.getTable(TEST_DB, TEST_TABLE)).thenReturn(hiveTable);
+    when(hiveTable.getPartitionKeys()).thenReturn(Lists.newArrayList());
```
max-jacobs: Just missing the case where the list of partition keys returned is null, but perhaps thats not even possible? probably fine to leave this out.

shunting-yard-replicator/src/main/java/com/expediagroup/shuntingyard/replicator/exec/messaging/MessageReaderAdapter.java
abhimanyugupta07: Hmm.. not sure. I have merged first and second `if` conditions into one. Looks better ?

shunting-yard-replicator/src/main/java/com/expediagroup/shuntingyard/replicator/exec/messaging/MessageReaderAdapter.java
abhimanyugupta07: Yes. Patrick also mentioned this. I have changed to `isNotEmpty(..)`

shunting-yard-replicator/src/test/java/com/expediagroup/shuntingyard/replicator/exec/messaging/MessageReaderAdapterTest.java
patduin: I think it can that's why I suggested `sourceTable.getPartitionKeysSize() > 0;` as that does the null check in the table class.

shunting-yard-replicator/src/main/java/com/expediagroup/shuntingyard/replicator/exec/messaging/MessageReaderAdapter.java
max-jacobs: Could make a method out of that one, `if(isSameLocation(...))`? But yep I'm happy with that, looks better.

shunting-yard-replicator/src/test/java/com/expediagroup/shuntingyard/replicator/exec/messaging/MessageReaderAdapterTest.java
abhimanyugupta07: It doesn't return `null`. It returns an empty list but I will still add a test case.

shunting-yard-replicator/src/main/java/com/expediagroup/shuntingyard/replicator/exec/context/CommonBeans.java
abhimanyugupta07: Yeah. We need two HiveConf, one each for source and target. It resolves by the method name I think.

PR #49, opened 2020-03-30
Changing LOG LEVEL for shunting-yard

abhimanyugupta07:


shunting-yard-binary/src/main/resources/log4j.xml
```diff
@@ -28,7 +28,7 @@
   </logger>
 
   <root>
-    <level value="INFO"/>
+    <level value="ERROR"/>
```
massdosage: I can see how this is the quickest and easiest way to reduce the amount of logging and I don't really mind having this as the default. My only question is whether this would be better done when ShuntingYard is deployed somewhere? For example in the Terraform or whatever, it could do a replace on this value and set it to whatever it wants it to be so could be customised per deployment use case. We could also provide a logging config template file which could be easier to do variable substitution on.

abhimanyugupta07: Yeah. That could be a good idea. Let me see what I can do and I will report back.

PR #52, opened 2020-09-23
egdl-1429: update parent 

shermosa:
These two PRs need to be merged first:
https://github.com/ExpediaGroup/eg-oss-plugin-config/pull/5
https://github.com/ExpediaGroup/eg-oss-parent/pull/11

- Changed the **parent** of shunting-yard to be eg-oss-parent (which also has the latest plugin versions) and not hotels-oss-parent.
- Changed the version from 3.1.1-SNAPSHOT to 3.2.0-SNAPSHOT
- In version 3.1.1 of shunting-yard there was an update in the hive version, but this was never released so it will be implemented in this release too (see changelog).


massdosage (approved):
Looks good, we'll wait on the parent release and update that here before merging.

abhimanyugupta07 (approved):

PR #54, opened 2020-09-25
egdl-1430: update plugins + replace cobertura for jacoco

shermosa:
[Ticket egdl-1430](https://jira.expedia.biz/browse/EGDL-1430).
- Update some plugins to specific versions.
- Cobertura removed and replaced with Jacoco for coverage (Jacoco is included in eg-oss-parent).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) a
 ### Changed
 - Upgraded shunting-yard parent from `hotels-oss-parent` to `eg-oss-parent`.
 - Upgraded version of `hive.version` to `2.3.7` (was `2.3.4`). Allows Shunting Yard to be used on JDK>=9.
+- Cobertura removed and replaced with Jacoco for coverage.
+- Upgraded `common-hive-metastore` from  1.4.2
 
 ## [3.1.0] - 2020-01-16
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) a
 ### Changed
 - Upgraded shunting-yard parent from `hotels-oss-parent` to `eg-oss-parent`.
 - Upgraded version of `hive.version` to `2.3.7` (was `2.3.4`). Allows Shunting Yard to be used on JDK>=9.
-- Cobertura removed and replaced with Jacoco for coverage.
 - Upgraded `hcommon.hive.metastore` from `1.4.1` to `1.4.2`.
 - Upgraded `circus.train` from `15.0.0` to `16.3.0`.
-- Upgraded `beeju.version` from `1.2.0` to `3.1.0`.
-- Upgraded `mockito.version` from `2.10.0` to `3.5.11`.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) a
 ### Changed
 - Upgraded shunting-yard parent from `hotels-oss-parent` to `eg-oss-parent`.
 - Upgraded version of `hive.version` to `2.3.7` (was `2.3.4`). Allows Shunting Yard to be used on JDK>=9.
+- Cobertura removed and replaced with Jacoco for coverage.
+- Upgraded `hcommon.hive.metastore` from `1.4.1` to `1.4.2`.
+- Upgraded `circus.train` from `15.0.0` to `16.3.0`.
+- Upgraded `beeju.version` from `1.2.0` to `3.1.0`.
+- Upgraded `mockito.version` from `2.10.0` to `3.5.11`.
```
massdosage, line 14: We normally don't call out changes to our unit testing dependencies in the changelog for our apps as the end users of the apps generally don't care about these. So you can remove these 2 from there. It's different with the parent POMs as the end users of those are developers and the would care about testing library and plugin versions and changes.

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) a
 ### Changed
 - Upgraded shunting-yard parent from `hotels-oss-parent` to `eg-oss-parent`.
 - Upgraded version of `hive.version` to `2.3.7` (was `2.3.4`). Allows Shunting Yard to be used on JDK>=9.
+- Cobertura removed and replaced with Jacoco for coverage.
```
massdosage, line 10: See below, no need to call this out in the CHANGELOG.

massdosage (changes requested):
POM changes look good, just a few comments from me on the CHANGELOG.

massdosage (approved):

pom.xml
```diff
@@ -30,29 +30,27 @@
...
-    <hcommon.hive.metastore.version>1.4.1</hcommon.hive.metastore.version>
+    <hcommon.hive.metastore.version>1.4.2</hcommon.hive.metastore.version>
     <derby.version>10.10.2.0</derby.version>
     <kafka.version>0.11.0.1</kafka.version>
-    <aws.version>1.11.415</aws.version>
+    <aws.version>1.11.852</aws.version>
     <aws.kcl.version>1.9.0</aws.kcl.version>
     <aws.kpl.version>0.12.8</aws.kpl.version>
-    <circus.train.version>15.0.0</circus.train.version>
+    <circus.train.version>16.3.0</circus.train.version>
```
patduin, line 43: this is a major version this won't affect SY? @abhimanyugupta07

massdosage: I looked through the release notes, the major version change in CT was due to Java 7 to 8 but SY is already Java 8 so that's fine. I don't think the rest has a direct impact on SY but as always appreciate another set of eyes. Everything compiles which is good ;)

patduin (approved):

pom.xml
abhimanyugupta07: Yeah. I don't think it will break anything. 🤞

abhimanyugupta07 (approved):

pom.xml
massdosage: I'll merge now.

